Issue #4, opened 2021-03-10
Use cases diagram


HellHunterMax:
# Tasks:
- [ ] Define use cases for website
- [ ] Describe use cases
- [ ] Describe roles
- [ ] Create Diagram

# Features & Use Cases:
### Admin
- Manage Jokes
- Manage Users
### Registered user.
- Favorite Jokes
- Like Joke
- CRUD jokes
- Add or +1 Tags on other people's jokes
### Guest
- Register
- View Jokes
- Report Joke
- Sort jokes
  - By type
  - By upload date
  - By Likes

# How To:
https://online.visual-paradigm.com/diagrams/tutorials/use-case-diagram-tutorial/

Make a pull request for the dev branch. and place all documents in `root/Documents/UML/`

PR #7, opened 2021-03-12
#4/docs/use cases diagram

HellHunterMax:
For issue #4  I created a file for the first 3 Items.

This was mostly a thinking progress for how people will use the website.
Normally I think this would be even more extensive but I believe this will serve its purpose here and does not need more time spent on it.
If you think there is anything missing pleas Edit it yourself of Review and comment here!


Ribby-0254 (approved):
Excellent work! 💯

PR #32, opened 2021-03-18
Feature 26 data context

### LINKED ISSUE #26: create DbContext
I Would say The DBContext should go in Data folder;
but it could be in the Helper folder? your choice.

Should be called : DataContext ?

Docs: 
- https://www.youtube.com/watch?v=fmvcAzHpsk8
- https://www.pointblankdevelopment.com.au/blog/134/aspnet-core-20-angular-24-user-registration-and-login-tutorial-example#data-context-cs

chrisK00:
close #26 

Created datacontext and db. Supressed xml warnings (i had accidently set them to information last time which changed the hidden file .editorconfig)

https://docs.microsoft.com/en-us/dotnet/api/system.data.entity.dbcontext?view=entity-framework-6.2.0


HellHunterMax (approved):
Seems good. I don't know if i like the service Extensions yet. but we will find out, we don't work with it that much so should be fine.

chrisK00:
Wrong branch

PR #45, opened 2021-03-19
Added more info to the README

HellHunterMax:
added info to Getting started and built with sections.
Still not finished with README though.


JBSE (approved):
Changes look good

PR #51, opened 2021-03-21
Added comments, fixed problems pointed out in comments, removed Newto…

N1c0z:
Choose everything from DB-Data-SeedingV2 branch since it's the most recent


Server/JokesOnYou.Web.Api/Data/DataContext.cs
```diff
@@ -15,5 +15,6 @@ public DataContext(DbContextOptions<DataContext> options) : base(options)
         public DbSet<User> Users { get; set; }
         public DbSet<Joke> Jokes { get; set; }
         //For each class in models folder, add it here, only if its also in db
+        // This is what was written in the wiki by ChrisK
```
chrisK00, line 18: you can remove this comment

Server/JokesOnYou.Web.Api/Data/DataSeeding.cs
```diff
@@ -15,18 +15,20 @@ public static void Initialize(DataContext context)
         {
             context.Database.EnsureCreated(); //Make sure db actually exists
 
-            // Look for any students.
+            // Look for any users in db
             if (context.Users.Any())
             {
+                //If there are students then leave method
```
chrisK00, line 21: you can remove this comment or if you want to keep it change students to users

Server/JokesOnYou.Web.Api/Data/DataSeeding.cs
```diff
@@ -15,18 +15,20 @@ public static void Initialize(DataContext context)
...
                 return;
             }
 
-            var jsonThingy = JsonConvert.DeserializeObject<List<User>>(File.ReadAllText(AppContext.BaseDirectory + "SeedingDB.json"));
+            //If you go here than it means that the db is empty
+            var TemporaryUsers = JsonSerializer.Deserialize<List<User>>(File.ReadAllText(@"Data\UserSeedData.json"));
```
chrisK00, line 26: should be camelcase
i would vote for it being named just users since its a list of users. Or usersToAdd

Server/JokesOnYou.Web.Api/Data/UserSeedData.json
```diff
@@ -0,0 +1,132 @@
...
+    "passwordhash": [
+      146,
+      63,
+      178,
+      108,
+      57,
+      149,
+      38
+    ],
+    "Passwordsalt": [
```
chrisK00, line 117: did the seeding work? considering in c# our props are capitalized and this is the only capitalized prop, and i didnt see you specifying options to ignore casings but it would probably be easier if the props were pascal case here instead.

another thing to consider if we are not gonna use microsoft identity (its a bit unclear if we are @HellHunterMax) we would want to hash and salt this right after we fetch the users so every user would have the same password before adding to the database. So when we are testing we use the same password.

Server/JokesOnYou.Web.Api/JokesOnYou.Web.Api.csproj
```diff
@@ -5,26 +5,19 @@
...
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
-    <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
     <PackageReference Include="Swashbuckle.AspNetCore" Version="6.1.1" />
-  </ItemGroup>
-
-  <ItemGroup>
-    <None Update="JokesOnYou.db">
-      <CopyToOutputDirectory>Never</CopyToOutputDirectory>
```
chrisK00, line 26: was this changed to copy if newer or just deleted?

chrisK00 (changes requested):

Server/JokesOnYou.Web.Api/Data/DataSeeding.cs
N1c0z: idk why but i keep writing students instead of users, xd

Server/JokesOnYou.Web.Api/Data/DataSeeding.cs
N1c0z: Oh, yeah should fix that, whoops

Server/JokesOnYou.Web.Api/Data/UserSeedData.json
N1c0z: Ok, will change the json file accordingly, so every user in json has the same password and password salt?

Server/JokesOnYou.Web.Api/JokesOnYou.Web.Api.csproj
N1c0z: There's no need to copy it anywhere? even if it gets deleted the program will generate a new db and put users inn it

chrisK00: its to avoid issues in the future, its very common. for example when publishing an app to a folder we need it to get the db.

copy paste from google:

`Copy always: This option works best when a developer intent is to have a clean slate to ensure their code functions correctly rather than having to clean up data on each build.

Copy if Newer: Use this option when the intent is to ensure changes to a database or a file persist between builds. Unlike [Copy always], [Copy if Newer] will leave files untouched unless there is a change to the file. For example, adding a new field or changing the field type in a database table will trigger a onetime copy, the same for working with any file, make a change and there will be a onetime copy.

This also means any data in a file within the build folder will be overwritten.

Do not Copy: As the option indicates, the file will not be copied. Note that changing from one of the other options will cause the file to be removed from the build folder. If the file in the build folder is needed then first, create a copy of the file before changing Copy to Output Directory to [Do not Copy].`

HellHunterMax:
Could we get some nfsw true's ? in the seed file ?

chrisK00 (approved):

chrisK00:
Looks perfect! Well done

PR #65, opened 2021-03-28
Feature/auth and roles

HellHunterMax:
THIS WILL (probably) NOT BE MERGED WITH DEV THIS IS JUST FOR REFERENCE.
I WILL ADD MORE INFO AND DOCUMENTATION TO THIS TONIGHT!

### DIFF
--- a/Server/JokesOnYou.Web.Api/Controllers/AuthController.cs
+++ b/Server/JokesOnYou.Web.Api/Controllers/AuthController.cs
@@ -24,12 +24,12 @@ namespace JokesOnYou.Web.Api.Controllers
 
         [AllowAnonymous]
         [HttpPost("login")]
-        public async Task<IActionResult> Login(UserLoginDTO userLoginDto)
+        public async Task<ActionResult<UserLoginDTO>> Login(UserLoginDTO userLoginDto)
         {
             try
             {
-                var user = await _authService.LoginAsync(userLoginDto);
-                return Ok(user);
+                var userDto = await _authService.LoginAsync(userLoginDto);
+                return Ok(userDto);
             }
             catch (UserLoginException ex)
             {
--- a/Server/JokesOnYou.Web.Api/Exceptions/AppException.cs
+++ b/Server/JokesOnYou.Web.Api/Exceptions/AppException.cs
@@ -1,0 +1,15 @@
+﻿using System;
+
+namespace JokesOnYou.Web.Api.Exceptions
+{
+    public class AppException : Exception
+    {
+        public AppException() : base()
+        {
+        }
+
+        public AppException(string message) : base(message)
+        {
+        }
+    }
+}
--- a/Server/JokesOnYou.Web.Api/Startup.cs
+++ b/Server/JokesOnYou.Web.Api/Startup.cs
@@ -36,11 +36,14 @@ namespace JokesOnYou.Web.Api
             services.AddScoped<IUserRepository, UserRepository>();
             services.AddScoped<IUnitOfWork, UnitOfWork>();
 
+            //This part is not needed now because there is no loop, it fixed the loop last time i encouterd it because Identity User has a SignUpDate loop.
+            /*
             services.AddMvc()
             .AddJsonOptions(o => {
                 o.JsonSerializerOptions
                     .ReferenceHandler = ReferenceHandler.Preserve;
             });
+            */
 
             services.AddSwaggerGen(config =>
             {
--- a/Tests/Server/UserServiceTests.cs
+++ b/Tests/Server/UserServiceTests.cs
@@ -1,0 +1,55 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+using FluentAssertions;
+using JokesOnYou.Web.Api.Repositories.Interfaces;
+using JokesOnYou.Web.Api.Services;
+using Moq;
+using JokesOnYou.Web.Api.Repositories;
+using JokesOnYou.Web.Api.Models;
+
+namespace Tests.Server
+{
+    public class UserServiceTests
+    {
+        private readonly Mock<IUserRepository> _Repo;
+
+        public UserServiceTests()
+        {
+            //_Repo = new UserService();
+        }
+
+        [Fact]
+        public void GetUserByID_Given_ExistingId_Returns_User()
+        {
+            var repo = new Mock<IUserRepository>();
+            var service = new UserService(repo.Object);
+            User ExpectedUser = new User() { Id = "1", Email = "a@a.a" };
+            repo.Setup(r => r.GetUserAsync("1").Result).Returns(ExpectedUser);
+
+            var actualUser = service.GetUserById("1").Result;
+
+            actualUser.Should().Be(ExpectedUser);
+
+        }
+
+        [Fact]
+        public void GetUserById_Given_NonExistingId_Throws()
+        {
+            //Arrange
+            var repo = new Mock<IUserRepository>();
+            var service = new UserService(repo.Object);
+            User ExpectedUser = new User();
+            repo.Setup(r => r.GetUserAsync("").Result).Returns((User)null);
+
+            //Act
+            Func<Task> act = async () => { await service.GetUserById(""); };
+
+            //Assert
+            act.Should().ThrowAsync<Exception>().WithMessage($"No user found with id: 1");
+        }
+    }
+}
--- a/Tests/Tests.csproj
+++ b/Tests/Tests.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>net5.0</TargetFramework>


Server/JokesOnYou.Web.Api/Controllers/AuthController.cs
```diff
@@ -7,16 +10,53 @@
...
+                return BadRequest(ex.Message);
+            }
+            /*
+            catch (AggregateException ex)
+            {
+                return BadRequest(ex.Message);
+            }*/
+        }
+
+        [AllowAnonymous]
```
chrisK00, line 45: should work without the allow anonymous since its default

Server/JokesOnYou.Web.Api/Controllers/AuthController.cs
```diff
@@ -7,16 +10,53 @@
...
+        [HttpPost("login")]
+        public async Task<IActionResult> Login(UserLoginDTO userLoginDto)
         {
-            throw new NotImplementedException();
+            try
+            {
+                var user = await _authService.LoginAsync(userLoginDto);
+                return Ok(user);
+            }
+            catch (UserLoginException ex)
```
chrisK00, line 34: we should setup the global exception handler, so good idea to save possible exceptions.

Server/JokesOnYou.Web.Api/Controllers/AuthController.cs
```diff
@@ -7,16 +10,53 @@
...
+        public AuthController(IAuthService authService)
         {
-            throw new NotImplementedException();
+            _authService = authService;
         }
 
-        public Task<ActionResult> Register([FromBody] UserRegisterDTO userRegisterDto)
+        [AllowAnonymous]
+        [HttpPost("login")]
+        public async Task<IActionResult> Login(UserLoginDTO userLoginDto)
```
chrisK00: specify that a user will be sent back

Server/JokesOnYou.Web.Api/Controllers/AuthController.cs
```diff
@@ -7,16 +10,53 @@
...
         {
-            throw new NotImplementedException();
+            try
+            {
+                var user = await _authService.LoginAsync(userLoginDto);
+                return Ok(user);
+            }
+            catch (UserLoginException ex)
+            {
+                return BadRequest(ex.Message);
```
chrisK00: return Unauthorized

Server/JokesOnYou.Web.Api/Extensions/ServiceExtensions.cs
```diff
@@ -18,8 +21,29 @@ public static IServiceCollection ConfigureAppServices(this IServiceCollection se
             services.AddDbContext<DataContext>(options =>
                 options.UseSqlite(config.GetConnectionString("DefaultConnection")));
 
-            services.AddIdentity<User, IdentityRole>().AddEntityFrameworkStores<DataContext>();
+            services.AddIdentity<User, IdentityRole>(
+                options => { options.SignIn.RequireConfirmedAccount = false; options.User.RequireUniqueEmail = true; }
```
chrisK00: sign in requireconfirmed account is fale by default i believe. Might wanna check if same applies to requniqmail

Server/JokesOnYou.Web.Api/Repositories/UserRepository.cs
```diff
@@ -16,6 +18,26 @@ public UserRepository(UserManager<User> userManager)
...
+            if (userCreationResult.Succeeded)
+            {
+                return await _userManager.FindByEmailAsync(userRegisterDTO.Email);
+            }
+            else
+            {
+                var message = string.Empty;
+                foreach (var error in userCreationResult.Errors)
+                {
+                    message += error.Description+ System.Environment.NewLine;
```
chrisK00: This is some nice validation, maybe wanna use stringbuilder for performance

Server/JokesOnYou.Web.Api/Repositories/UserRepository.cs
```diff
@@ -27,6 +49,11 @@ public async Task<User> GetUserAsync(string id)
             return user;
```
chrisK00: shouldnt return user if he is deleted

Server/JokesOnYou.Web.Api/Services/AuthService.cs
```diff
@@ -1,22 +1,60 @@
...
+            var result = await _signInManager.CheckPasswordSignInAsync(user, userLoginDTO.Password, false);
+
+            //var result = await _signInManager.PasswordSignInAsync(userLoginDTO.Email, userLoginDTO.Password, false, false);
+
+            if (result.Succeeded)
+            {
+                //TODO need to assign user a role!
+                UserReplyDTO userReplyDTO = new UserReplyDTO() { Email = user.Email, Id = user.Id, UserName = user.UserName, Role = "Registered" };
+
+                string token = _tokenService.GetToken(user);
```
chrisK00: for some more decoupling, maybe our token creator should just recieve what it needs, like the roles and username

Server/JokesOnYou.Web.Api/Services/JwtTokenService.cs
```diff
@@ -1,17 +1,62 @@
...
         }
+
+
+        private string GenerateToken(User user)
+        {
+            // generate token that is valid for 7 days
+            var tokenDescriptor = new SecurityTokenDescriptor
+            {
+                Subject = new ClaimsIdentity(new[] { new Claim(ClaimTypes.Name, user.Id), new Claim(ClaimTypes.Role, "Registered") }),
+                Expires = DateTime.UtcNow.AddDays(7),
```
chrisK00: we can set this to a day

chrisK00 (commented):
just a couple of thoughts. Well done overall! Looks great.

Server/JokesOnYou.Web.Api/Controllers/AuthController.cs
HellHunterMax: Ah yes you are right I forgot to add the [Authorize] attribute. I want the Writer of the code to be explicit in his commands so when something should be accessible to everyone it should be [AllowAnonymous]. for the case of adding a new endpoint as well.

I was also reading about DefaultIdentity There is defaults to authorize everything. that got me mixed up as well.

But either way would work, removing it or adding authorize.

Server/JokesOnYou.Web.Api/Controllers/AuthController.cs
HellHunterMax: Yes we should I need to read up on global exception handlers ;) though this is not an unhandled exception so this should not give an issue and I think is good practice.

Server/JokesOnYou.Web.Api/Controllers/AuthController.cs
HellHunterMax: How ?

Server/JokesOnYou.Web.Api/Controllers/AuthController.cs
HellHunterMax: Oké so 401 it is ;)

`**400 Bad Request**`
The server cannot or will not process the request due to an apparent client error (e.g., malformed request syntax, size too large, invalid request message framing, or deceptive request routing).
`**401 Unauthorized**`
Similar to 403 Forbidden, but specifically for use when authentication is required and has failed or has not yet been provided.

It seems to be the standard.

source: https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_client_errors

Server/JokesOnYou.Web.Api/Extensions/ServiceExtensions.cs
HellHunterMax: Property                         | Description                                           | Default
RequireConfirmedEmail | Requires a confirmed email to sign in. | false


Property                   | Description                                                    | Default
RequireUniqueEmail | Requires each user to have a unique email. | false

Server/JokesOnYou.Web.Api/Repositories/UserRepository.cs
HellHunterMax: ah this is a weird one in Github. there are some lines missing above here. You can expand to see the missing lines.
DeleteUserAsync does not return a user :P
But you made me look at Looking up users and found a null error ;)

Server/JokesOnYou.Web.Api/Services/AuthService.cs
HellHunterMax: I'm not so sure about this one.
We could but would mean more work sometimes.
```
Every time we want to add anything to the GetToken we have to change this method and the GetToken.
If we add more places where GetToken gets called, they all have to change their parameters when GetToken changes.
```
```
Now if we add stuff to the Token we don't need to change this nor the parameters.
```
So i think this is better. But please do convince me!

Server/JokesOnYou.Web.Api/Services/JwtTokenService.cs
HellHunterMax: yes we could do anything with it :D

Server/JokesOnYou.Web.Api/Controllers/AuthController.cs
chrisK00: Change the iactionresult to ActionResult<User. Remove the Ok() that wraps the user. This will give us type safety and an Ok response is default when sending an item back i think

HellHunterMax: image
that didnt work but this did:
image

PR #66, opened 2021-03-29
Feature/adding interfaces

HellHunterMax:
I think this skeleton should be good. now everyone should be able to work on their separate issues.
Lets merge this as quickly as possible so people can start working on this new revised version.


chrisK00 (approved):

JBSE (approved):

HellHunterMax:
I'll make 1 change to return type of the login Controller method.

PR #69, opened 2021-04-05
created appexception

### LINKED ISSUE #68: create appexception class for bad requests

chrisK00:
close #68

### DIFF
--- a/Server/JokesOnYou.Web.Api/Exceptions/AppException.cs
+++ b/Server/JokesOnYou.Web.Api/Exceptions/AppException.cs
@@ -4,7 +4,7 @@ namespace JokesOnYou.Web.Api.Exceptions
 {
     public class AppException : Exception
     {
-        public AppException() : base()
+        public AppException()
         {
         }
 
--- a/Server/JokesOnYou.Web.Api/Exceptions/AppException.cs
+++ b/Server/JokesOnYou.Web.Api/Exceptions/AppException.cs
@@ -1,0 +1,15 @@
+﻿using System;
+
+namespace JokesOnYou.Web.Api.Exceptions
+{
+    public class AppException : Exception
+    {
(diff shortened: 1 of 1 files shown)


Server/JokesOnYou.Web.Api/Exceptions/AppException.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+
+namespace JokesOnYou.Web.Api.Exceptions
+{
+    public class AppException : Exception
+    {
+        public AppException() : base()
```
HellHunterMax: don't think the base here is needed ?
https://docs.microsoft.com/en-us/dotnet/standard/exceptions/how-to-create-user-defined-exceptions

HellHunterMax (approved):
Just wondering if the inner exception would be handy but we can always add it later if we need it.

chrisK00:
> Just wondering if the inner exception would be handy but we can always add it later if we need it.

Its just hard for me think about. Like imagine a user field is null, we throw a KeyNotFoundException, our middleware catches it and sends back a 404 not found. When would we add more exceptions and how would that be sent back? To me it only makes sense in validation since your validating multiple stuff. If you have an example id love that.

Server/JokesOnYou.Web.Api/Exceptions/AppException.cs
chrisK00: Correct. Its a matter of inheritance and usage of base keyword tho.

HellHunterMax:
For example: the user register method returns with multiple errors inside an aggregate exception. You could add the aggregate exception as inner exception. Not sure if it's good but it's possible. 

I really don't know if we will need it. We will see when we get to that point and can always then add it.

code-reaper08 (approved):

PR #70, opened 2021-04-06
fix: remove DeleteUserAsync(int id) from IUserRepository

HellHunterMax:
The id is not an int and we already have DeleteUserAsync(User user);


ofu997 (commented):
makes sense

ofu997 (approved):
approve

chrisK00 (approved):

PR #96, opened 2021-04-07
Feature/92 global ex handler

### LINKED ISSUE #92: Feature: Add Global exception handler
add a global exception handler to keep controllers lean;

chrisK00:
Created a simple global exception handler that will catch any errors in our application. Id consider using built exceptions instead of creating lots of multiple custom in the future. F.e UserRegisterException can be an argument exception


JBSE (approved):
looks good

HellHunterMax (approved):

PR #99, opened 2021-04-07
57 feature create account controller2

### LINKED ISSUE #76: Endpoint: Register User
Create the whole line for Register User API call.

- [ ] RegisterUser inside controller HTTP post
- [ ] RegisterUser inside Service
- [ ] RegisterUser inside Repository

Use UserManager from identity.

### LINKED ISSUE #77: Endpoint: Login User
Create the whole line for Login User API call.

- [ ] LoginUser inside controller HTTP post
- [ ] LoginUser inside Service
- [ ] LoginUser inside Repository

<details>
  <summary>How To / tips:</summary>

- use ITokenService to get the token.

- Use UserManager from identity to get the user.

- Use SignInManager from identity to check sign in.

- Return UserReplyDto.
</details>

ofu997:
attempting #76 register and #77 login


HellHunterMax (changes requested):
So like i told you on Discord.
You can try to split this up into, controller, Service and Repository.

- The Controller just takes the call instructs other classes and returns.
- Move all the "work" into the service.
- The Repository makes the Connection to the Database.

Server/JokesOnYou.Web.Api/Controllers/AccountController.cs
```diff
@@ -0,0 +1,102 @@
...
+    [ApiController]
+    [Authorize]
+    public class AccountController : ControllerBase
+    {
+        private readonly UserManager<User> _userManager;
+        private readonly SignInManager<User> _signInManager;
+        private readonly ITokenService _ITokenService;
+        private readonly IUserRepository _IUserRepository;
+
+        public AccountController(
```
chrisK00: Very subjective opinion but id have theese on prolly ~2 lines since they are not really that important

Server/JokesOnYou.Web.Api/Controllers/AccountController.cs
```diff
@@ -0,0 +1,102 @@
...
+namespace JokesOnYou.Web.Api.Controllers
+{
+    [Route("api/[controller]")]
+    [ApiController]
+    [Authorize]
+    public class AccountController : ControllerBase
+    {
+        private readonly UserManager<User> _userManager;
+        private readonly SignInManager<User> _signInManager;
+        private readonly ITokenService _ITokenService;
```
chrisK00: All of theese dont belong here, service should make use of them

Server/JokesOnYou.Web.Api/Controllers/AccountController.cs
```diff
@@ -0,0 +1,102 @@
...
+using JokesOnYou.Web.Api.Data;
+using JokesOnYou.Web.Api.Services.Interfaces;
+using JokesOnYou.Web.Api.DTOs;
+using JokesOnYou.Web.Api.Repositories.Interfaces;
+
+namespace JokesOnYou.Web.Api.Controllers
+{
+    [Route("api/[controller]")]
+    [ApiController]
+    [Authorize]
```
chrisK00: This api controller does not need [auth] considering its for login/register anyone is able to, instead of adding [allowAnonymous] on every method

Server/JokesOnYou.Web.Api/Controllers/AccountController.cs
```diff
@@ -0,0 +1,102 @@
...
+            _ITokenService = ITokenService;
+            _IUserRepository = IUserRepository;
+        }
+
+        [AllowAnonymous]
+        [HttpPost("login")]
+        public async Task<ActionResult> Login (UserLoginDTO userLogin)
+        {
+            // If user has registered
+            var user = await _IUserRepository.GetUserByEmail(userLogin.Email);
```
chrisK00: All logic should be inside of the service. Message me when youve edited the code @N1c0z commented about

Server/JokesOnYou.Web.Api/Controllers/AccountController.cs
```diff
@@ -0,0 +1,102 @@
...
+            {
+                return Unauthorized("Sorry, we could not find an account with that email");
+            }
+        }
+
+        [AllowAnonymous]
+        [HttpPost("register")]
+        public async Task<ActionResult> Register(UserRegisterDTO userRegister)
+        {
+            var result = await _IUserRepository.CreateUserAsync(userRegister);
```
chrisK00: Move to service. I think we might wanna do something with the result and not return it. We will return NoContent() if everything works as expected, otherwise the errors.

chrisK00 (changes requested):

Server/JokesOnYou.Web.Api/Services/UserService.cs
```diff
@@ -37,5 +44,40 @@ public async Task UpdateUser(UserUpdateDTO updateDTO)
...
+                var signInResult = await _signInManager.CheckPasswordSignInAsync(user, userLogin.Password, false);
+
+                if (signInResult.Succeeded)
+                {
+                    var userReplyDTO = new UserReplyDTO()
+                    {
+                        Id = user.Id,
+                        Email = user.Email,
+                        UserName = user.Email,
+                        Role = "",
```
Niisc: I still don't get why return `Role = ""`

Server/JokesOnYou.Web.Api/Services/UserService.cs
```diff
@@ -37,5 +44,40 @@ public async Task UpdateUser(UserUpdateDTO updateDTO)
...
+                        Token = _tokenService.GetToken(user)
+                    };
+                    return userReplyDTO;
+                }
+                else
+                {
+                    throw new NotImplementedException("Sign in failed");
```
Niisc: Return Unautorized instead?

Server/JokesOnYou.Web.Api/Services/UserService.cs
```diff
@@ -37,5 +44,40 @@ public async Task UpdateUser(UserUpdateDTO updateDTO)
...
+                }
+            }
+            else
+            {
+                throw new NotImplementedException("User not found");
```
Niisc: Return Unautorized instead?

HellHunterMax: Service should not return any http, so throwing the exception is good but I would change it to AppException because it's not not implemented. The global exception handler will handle the return.

Server/JokesOnYou.Web.Api/Services/UserService.cs
ofu997: Can i just delete the role line? Not sure which roles we have defined so far

Server/JokesOnYou.Web.Api/Controllers/AccountController.cs
```diff
@@ -0,0 +1,47 @@
...
+        public AccountController(IUserService userService)
+        { 
+            _userService = userService;
+        }
+
+        [HttpPost("login")]
+        public async Task<ActionResult> Login (UserLoginDTO userLogin)
+        {
+            var loginResult = await _userService.LoginUser(userLogin);
+            return Ok(loginResult);
```
chrisK00: Good idea to specify the return type and then just return it so we know what we return+get type safety

Server/JokesOnYou.Web.Api/Controllers/AccountController.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            var loginResult = await _userService.LoginUser(userLogin);
+            return Ok(loginResult); 
+        }
+
+        [HttpPost("register")]
+        public async Task<ActionResult> Register(UserRegisterDTO userRegister)
+        {
+            var result = await _userService.RegisterUser(userRegister);
+            return Ok(result);
```
chrisK00: what is being sent back?

Server/JokesOnYou.Web.Api/Services/Interfaces/IUserService.cs
```diff
@@ -13,5 +15,7 @@ public interface IUserService
         Task<User> GetUserById(string id);
```
chrisK00, line 15: when do we return a complete user to the view? (user contains password etc...)

Server/JokesOnYou.Web.Api/Services/UserService.cs
chrisK00: Yeah you can just remove it, we arent handling roles right now

Server/JokesOnYou.Web.Api/Services/UserService.cs
```diff
@@ -37,5 +44,40 @@ public async Task UpdateUser(UserUpdateDTO updateDTO)
...
         }
+
+        public async Task<UserReplyDTO> LoginUser(UserLoginDTO userLogin)
+        {
+            var user = await _userRepository.GetUserByEmail(userLogin.Email);
+            if (user != null)
+            {
+                var signInResult = await _signInManager.CheckPasswordSignInAsync(user, userLogin.Password, false);
+
+                if (signInResult.Succeeded)
```
chrisK00: I would recommend always doing i think its called inverted if statements. So instead of checking if it succeeded, check if it failed and just throw an exception. Will keep the code clean

Server/JokesOnYou.Web.Api/Services/UserService.cs
chrisK00: > Return Unautorized instead?

since hes inside the service he cant return unauthorized but he can throw an unauthorizedexception and have the global exc handler catch it

Server/JokesOnYou.Web.Api/Services/UserService.cs
chrisK00: > Service should not return any http, so throwing the exception is good but I would change it to AppException because it's not not implemented. The global exception handler will handle the return.

Use the built in `throw new UnauthorizedAccessException();` then just add it to the ones being catched inside the global ex handler and return a 401 unauthorized

chrisK00 (changes requested):
Looks alot better now that the business logic is not in the same file 👍

Server/JokesOnYou.Web.Api/Controllers/AccountController.cs
ofu997: nothing

Server/JokesOnYou.Web.Api/Services/Interfaces/IUserService.cs
ofu997: changed to plain Task, so register only creates user.

HellHunterMax (approved):
All looks good now tested it and works.

Server/JokesOnYou.Web.Api/Controllers/AccountController.cs
```diff
@@ -23,25 +23,25 @@ namespace JokesOnYou.Web.Api.Controllers
...
         }
 
         [HttpPost("login")]
-        public async Task<ActionResult> Login (UserLoginDTO userLogin)
+        public async Task<ActionResult<UserReplyDTO>> Login (UserLoginDTO userLogin)
         {
-            var loginResult = await _userService.LoginUser(userLogin);
-            return Ok(loginResult); 
+            var userReplyDTO = await _userService.LoginUser(userLogin);
+            return Ok(userReplyDTO);
```
chrisK00: Since you specified the return type of the method it should be ok to skip the Ok() wrapper in order to provide type safety
- side note: doesnt work with ienumerables.

chrisK00 (approved):
Well done! The hard work seems to pay off 🥇

Server/JokesOnYou.Web.Api/Controllers/AccountController.cs
HellHunterMax: Oooh that's interesting now i see it, seems like last time i actually didn't get it ;). now i understand, thanks!

ofu997: removed wrapper

PR #101, opened 2021-04-12
split up the ViewedJokes in the architecture diagram.

### LINKED ISSUE #94: Update Models for the Split up ViewedJoke Model
Right now viewed joke is too big and will be used to much with too much useless information being sent
So we split up the ViewedJoke Into smaller models:

- [ ] Liked Joke
- [ ] Saved Joke
- [ ] Liked Tags (on joke)

update them in the Docs/Achitecture folder

HellHunterMax:
### What I did in this Branch:
1. split up the ViewedJokes in the architecture diagram into:
    - SavedJoke
    - LikedJoke
    - LikedTags

2. Also changed some connections because they were wrong.
3. Added the Architecture folder to the sln file, including the files in it.

### Why I did it:
I got told that having more small models is better for the database.
When we used 1 big class for viewedjokes we would get a lot of useless information every time we call it. now we can be very specific with the information we want to pull out of the Database.

**Old way:**
```
Want to get a list of saved jokes:
1. Database calls ViewedJokes 
2. filters out everything except the saved jokes.
3. Then filters out the useless info from theViewedjoke model into SavedJokesDto.
4. returns Dto's.
```
**New way:**
```
Want to get a list of saved jokes:
1. Database calls SavedJokes 
2. changes SavedJokes into SavedJokesDto
3. returns Dto's.
```

Also every time a Database is getting accessed its locked until its finished(its very fast but still). so having everything in 1 big table is not good.

closes #94


chrisK00 (commented):
I agree, good idea to split them up! Sent a message on discord to just clear up some things regarding the modelling

code-reaper08 (approved):

HellHunterMax:
Ill change the name of the ViewedDate tonight and remove the boolean in SavedJoke.

HellHunterMax:
Hmm might be handy to add IFlaggable to Flags so we can flag User, Joke and Tag in one place.
And a Tag needs a list of Jokes.
Ill update them again.

chrisK00 (approved):
Very good job on splitting the entities up!

PR #104, opened 2021-04-15
Feature/73 get all jokes

### LINKED ISSUE #73: Endpoint: Get all Jokes
Create the whole line for Get all Jokes API call.

- [x] GetJokes inside controller HTTP get
- [x] GetJokes inside JokeService
- [x] GetJokes inside JokeRepository

Return an IEnumerable<JokesReplyDto>

HellHunterMax:
# Created The endpoint for getting all jokes.

## How it works.
1. The Controller takes on the call. asks the Service to get it a list of all JokeReplyDto's.
2. The Service then asks the Repo for a List of all Jokes.
3. The repo gets all Jokes from the Database and includes a User (author) per Joke. ( that it gets all users might be changed in future ?) 
This is done by eager loading (telling the Database that you want to get the linked user.). 
4. Repo returns the list to the Service.
5. Service then maps the Jokes to a JokesReplyDto.
6. Service returns the list of JokesReplyDto.
7. Controller Returns the list to te caller.

## Issues I encounterd during this Quest:
1. I had to figure out mapping again.
2. Mapping the User into a Author inside the joke mapper.
3. Loading the User in the Joke from the DB.
4. not returning a whole user with the Joke.

## how i solved them:
1. I looked at @N1c0z code for the Mapper. and looked at [Automapper Documentation](https://docs.automapper.org/en/stable/Getting-started.html)
2. I [found a Stackoverflow post that helped](https://stackoverflow.com/questions/57257077/add-inner-mapper-to-automapper-mapping-profile)
Added this specific code:
```CreateMap<Joke, JokeReplyDto>()
                .ForPath(destination => destination.Author.Id, options =>
                    options.MapFrom(Source => Source.Author.Id))
                .ForPath(destination => destination.Author.UserName, options =>
                    options.MapFrom(source => source.Author.UserName));
```
3. [Different types of loading found here on MS docs](https://docs.microsoft.com/en-us/ef/core/querying/related-data/)
    [and found some explanation here c-shapcorner](https://www.c-sharpcorner.com/UploadFile/dacca2/lazy-and-eager-loading-in-entity-framework/)
4. Created the JokeAuthorDto to not return the whole user when returning a joke


code-reaper08 (approved):

ofu997:
tested this out and it works.

ofu997 (approved):

Server/JokesOnYou.Web.Api/Controllers/JokesController.cs
```diff
@@ -0,0 +1,33 @@
...
+        private readonly IJokesService _jokesService;
+
+        public JokesController(IJokesService jokesService)
+        {
+            _jokesService = jokesService;
+        }
+
+        [AllowAnonymous]
+        [HttpGet]
+        public async Task<ActionResult<IEnumerable<JokeReplyDto>>> GetAllJokesAsync()
```
chrisK00, line 26: Suggestion: No need to name the method async considering no one will be awaiting it

Server/JokesOnYou.Web.Api/JokesOnYou.Web.Api.csproj
```diff
@@ -14,6 +14,8 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="AutoMapper" Version="10.1.1" />
+    <PackageReference Include="AutoMapper.Extensions.Microsoft.DependencyInjection" Version="8.1.1" />
```
chrisK00, line 17: we only need this package

Server/JokesOnYou.Web.Api/Profiles/JokesProfile.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Threading.Tasks;
+
+namespace JokesOnYou.Web.Api.Profiles
+{
+    public class JokesProfile : Profile
+    {
+        public JokesProfile()
+        {
+            CreateMap<Joke, JokeReplyDto>()
+                .ForPath(destination => destination.Author.Id, options =>
```
chrisK00: Theres no need to do forpath/config for every property. If we would have to do it for every entity/prop automapper becomes kinda useless. The way automapper works is it compares the property names in both entities, if they match it maps the one to the other.

Server/JokesOnYou.Web.Api/Services/JokesService.cs
```diff
@@ -0,0 +1,32 @@
...
+        public JokesService(IJokesRepository jokesRepo, IMapper mapper)
+        {
+            _jokesRepo = jokesRepo;
+            _mapper = mapper;
+        }
+
+        public async Task<IEnumerable<JokeReplyDto>> GetAllJokesAsync()
+        {
+            var jokes = await _jokesRepo.GetAllJokesAsync();
+
```
chrisK00, line 26: Probably the best way to handle this is to use automapper https://docs.automapper.org/en/stable/Queryable-Extensions.html. This way we only SELECT the properties we need from the db and we dont even need to manually include navigation props.

Server/JokesOnYou.Web.Api/Services/JokesService.cs
```diff
@@ -0,0 +1,32 @@
...
+        private readonly IJokesRepository _jokesRepo;
+        private readonly IMapper _mapper;
+
+        public JokesService(IJokesRepository jokesRepo, IMapper mapper)
+        {
+            _jokesRepo = jokesRepo;
+            _mapper = mapper;
+        }
+
+        public async Task<IEnumerable<JokeReplyDto>> GetAllJokesAsync()
```
chrisK00: Suggestion: GetAllJokesDtosAsync

Server/JokesOnYou.Web.Api/Startup.cs
```diff
@@ -36,13 +37,17 @@ public void ConfigureServices(IServiceCollection services)
             services.AddScoped<ITokenService, JwtTokenService>();
             services.AddScoped<IUserRepository, UserRepository>();
             services.AddScoped<IUnitOfWork, UnitOfWork>();
+            services.AddScoped<IJokesRepository, JokesRepository>();
+            services.AddScoped<IJokesService, JokesService>();
 
             services.AddMvc()
             .AddJsonOptions(o => {
```
chrisK00: not sure when i missed this, could you explain? (if you wrote it)

Tests/Server/JokesServiceTests.cs
```diff
@@ -0,0 +1,77 @@
...
+{
+    public class JokesServiceTests
+    {
+        private readonly IMapper _mapper = new Mapper(new MapperConfiguration(cfg => cfg.AddProfile<JokesProfile>()));
+
+        [Fact]
+        public void GetAllUsers_Returns_ListOfJokesDto()
+        {
+            var repo = new Mock<IJokesRepository>();
+            var service = new JokesService(repo.Object, _mapper);
```
chrisK00: Use the constructor as a setup method, it will drastically reduce the amount of code in our tests

Tests/Server/JokesServiceTests.cs
```diff
@@ -0,0 +1,77 @@
...
+                new Joke() { Id = 2, Premise = "2", Punchline = "2", Likes = 2, Dislikes = 0}
+            };
+            repo.Setup(r => r.GetAllJokesAsync().Result).Returns(Jokes);
+            List<JokeReplyDto> expected = new List<JokeReplyDto>();
+            foreach (Joke joke in Jokes)
+            {
+                expected.Add(new JokeReplyDto() { Id = joke.Id, Premise = joke.Premise, Punchline = joke.Punchline, Likes = joke.Likes, Dislikes = joke.Dislikes });
+            }
+
+            var actualList = service.GetAllJokesAsync().Result;
```
chrisK00: How about awaiting this async method instead of trying to get the result out?

chrisK00 (changes requested):
Good job. I like that you dont instantly return all items in the controller, making it easier to debug.

Server/JokesOnYou.Web.Api/JokesOnYou.Web.Api.csproj
HellHunterMax: ok ill remove and test again ;)

Server/JokesOnYou.Web.Api/Profiles/JokesProfile.cs
HellHunterMax: ill try without. i thought i did try it without and didn't work but ill try again!

Server/JokesOnYou.Web.Api/Services/JokesService.cs
HellHunterMax: oh cool thanks! i check this out tonight!

Server/JokesOnYou.Web.Api/Startup.cs
HellHunterMax: woeps that should have been commented out in a fix earlier back hmm ill remove it again.

Tests/Server/JokesServiceTests.cs
HellHunterMax: https://stackoverflow.com/questions/29809302/is-there-a-difference-between-calling-result-or-await-on-known-completed-tasks#:~:text=The%20only%20functional%20difference%20that,will%20just%20raise%20the%20exception.

So i think not much difference but await needs the whole test to be async. but ill test it out anyway would like to see.

HellHunterMax:
Thanks for all the comments ill fix this tonight!

Tests/Server/JokesServiceTests.cs
chrisK00: > https://stackoverflow.com/questions/29809302/is-there-a-difference-between-calling-result-or-await-on-known-completed-tasks#:~:text=The%20only%20functional%20difference%20that,will%20just%20raise%20the%20exception.
> 
> So i think not much difference but await needs the whole test to be async. but ill test it out anyway would like to see.

actually theres more to it, thats why async/await exists. https://stackoverflow.com/questions/58312479/is-accessing-async-result-considered-as-bad-practice

I think theres a ReturnsAsync() method aswell for Moq so you should be able to figure it out np

Server/JokesOnYou.Web.Api/Controllers/JokesController.cs
HellHunterMax: There is no need indeed.
I'm just appending Async to all async methods no matter where to make it a habit for myself.
I'd rather have to much information then forgetting it somewhere. 

So I think I'll keep it but, you can change my mind of course.

Server/JokesOnYou.Web.Api/Profiles/JokesProfile.cs
HellHunterMax: Added 
```
```
Works now.

Server/JokesOnYou.Web.Api/Services/JokesService.cs
HellHunterMax: Very nice this works. exactly what I was worried about ;)
I tried different as well but that didn't work.
I tried 
```
var jokes = await _context.Jokes.Include( joke => joke.Author.Id)
                                .Include(joke => joke.Author.UserName)
                                .ToListAsync();
```
Must be a way to get this to work? but maybe not.

Tests/Server/JokesServiceTests.cs
HellHunterMax: So the Service doesn't do anything anymore. what should be tested is the mapper now ;)
removed the test.

Server/JokesOnYou.Web.Api/Repositories/JokesRepository.cs
```diff
@@ -0,0 +1,43 @@
...
+            var jokes = await _context.Jokes.Include( joke => joke.Author).ToListAsync();
+            return jokes;
+        }
+
+        public async Task<IEnumerable<JokeReplyDto>> GetAllJokeDtosAsync()
+        {
+            var configuration = new MapperConfiguration(cfg =>
+            {
+                cfg.CreateMap<Joke, JokeReplyDto>();
+                cfg.CreateMap<User, JokeAuthorDto>();
```
HellHunterMax: Any other way of doing this since i already have the mapping profile?

chrisK00: Yes. So since we have the profile configured, and automapper is added to DI serviceprovider. We can just inject the mapper using IMapper interface and use it all over the place.

chrisK00 (approved):
Inject automapper using IMapper and then this is ready for merge!

Server/JokesOnYou.Web.Api/Repositories/JokesRepository.cs
HellHunterMax: Very nice! 
_mapper.ConfigurationProvider worked here I was thinking there should be a way :D

PR #105, opened 2021-04-16
Feature/71 create joke

### LINKED ISSUE #71: Endpoint: Create Joke
Create the whole line for create joke API call.

- [x] CreateJoke inside controller HTTP post 
- [x] CreateJoke inside JokeService
- [x] CreateJoke inside JokeRepository

Joke should not be linked to a User yet!

HellHunterMax:
# This implements endpoint Create a joke.
closes #71.

### how it works:
1. Controller takes the call.
2. controller gets the User Id from the claim.
3. controller gives the JokeDto and User Id to the Service.
4. Service Gets the User from the UserRepo.
5. Service maps the JokeDto into a Joke.
6. service places the User into Joke.Author.
7. Service gives the joke to the JokesRepo.
8. JokesRepo places the Joke into DB.
9. Service asks UnitOfWork to save the DB.
10. Service Places the Joke into a JokeReplyDto
11. service returns the JokeReplyDto to the Controller
12. controller Returns the JokeReplyDto to the caller.

### issues I had on this:
1. getting the user Id without giving it though the model. we could mix both for extra safety?
2. how to save the Database.
3. Do i call the UserRepo inside the JokesService or should JokesService call the UserService?
4. should we call the database after saving and check if its in there ?

### how They were fixed:
1. Chris had a nice extension for this.
2. Saving is done through UnitOfWork, though still have question about it.
3. ???
4. ???

I have some questions check comments.


Server/JokesOnYou.Web.Api/Controllers/JokesController.cs
```diff
@@ -15,5 +21,24 @@ public JokesController(IJokesService jokesService)
         {
             _jokesService = jokesService;
         }
+
+        [Authorize(Roles = "registered,admin")]
+        [HttpPost]
+        public async Task<ActionResult<JokeReplyDto>> CreateJokeAsync(JokeCreateDto jokeCreateDto)
+        {
+
+            var userId = ClaimsPrincipalExtension.GetUserId(User);
```
HellHunterMax, line 30: is this a good way of getting the ID? We could combine it with the caller (website) actually supplying the User Id as well?
checking if both are the same ?

Server/JokesOnYou.Web.Api/Services/JokesService.cs
```diff
@@ -10,10 +14,42 @@ namespace JokesOnYou.Web.Api.Services
...
         {
             _jokesRepo = jokesRepo;
+            _mapper = mapper;
+            _userRepository = userRepository;
+            _unitOfWork = unitOfWork;
+        }
+        public async Task<JokeReplyDto> CreateJokeAsync(JokeCreateDto jokeCreateDto, string userId)
+        {
+
+            var user = await _userRepository.GetUserAsync(userId);
```
HellHunterMax, line 39: should this get the user like this or ask the UserService to get it ?
I'm guessing this is good.

Server/JokesOnYou.Web.Api/Services/JokesService.cs
```diff
@@ -10,10 +14,42 @@ namespace JokesOnYou.Web.Api.Services
...
+            var joke = _mapper.Map<Joke>(jokeCreateDto);
+            joke.UploadDate = DateTime.Now;
+            joke.Author = user;
+
+            await _jokesRepo.CreateJokeAsync(joke);
+            var saved = await _unitOfWork.SaveAsync();
+            if (!saved)
```
HellHunterMax, line 46: Should this check be in place because some other method could have saved it before this gets called ?

Server/JokesOnYou.Web.Api/Services/JokesService.cs
```diff
@@ -10,10 +14,42 @@ namespace JokesOnYou.Web.Api.Services
...
+            {
+                throw new AppException("Error with saving the Joke.");
+            }
+
+            var jokeReplyDto = _mapper.Map<JokeReplyDto>(joke);
```
HellHunterMax, line 51: Should we Get the Joke from the database to double check if its there?
Then return that instance ?

Server/JokesOnYou.Web.Api/Controllers/JokesController.cs
chrisK00: the token can be trusted, considering the key always stays on the server. Getting the userid from token ensures the person trying to do the action is the person with that token

Server/JokesOnYou.Web.Api/Models/Flag.cs
```diff
@@ -7,5 +8,10 @@ namespace JokesOnYou.Web.Api.Models
 {
     public class Flag
     {
+        public int Id { get; set; }
+        public Reason Reason { get; set; }
+        public User Issuer { get; set; }
+        public DateTime Created { get; set; }
```
chrisK00, line 14: Created will be = Datetime.Now considering we only create this entity, well when we add it

Server/JokesOnYou.Web.Api/Models/LikedJoke.cs
```diff
@@ -7,5 +7,10 @@ namespace JokesOnYou.Web.Api.Models
 {
     public class LikedJoke
     {
+        public int Id { get; set; }
+        public Joke Joke { get; set; }
+        public bool Liked { get; set; }
+        public DateTime LikeDate { get; set; }
+        public User User { get; set; }
```
chrisK00, line 14: I dont think a joke has a user, a user has a joke, it does have a userId though since it is owned by someone. Same with tags

Server/JokesOnYou.Web.Api/Models/Reason.cs
```diff
@@ -7,5 +7,8 @@ namespace JokesOnYou.Web.Api.Models
 {
     public class Reason
     {
+        public int Id { get; set; }
+        public string Name { get; set; }
+        public int Severity { get; set; }
```
chrisK00, line 12: Maybe we could use data annotations such as Range()

Server/JokesOnYou.Web.Api/Models/SavedJoke.cs
```diff
@@ -7,5 +7,9 @@ namespace JokesOnYou.Web.Api.Models
 {
     public class SavedJoke
     {
+        public int Id { get; set; }
+        public Joke Joke { get; set; }
+        public DateTime SavedDate { get; set; }
```
chrisK00, line 12: Same as the created date comment

Server/JokesOnYou.Web.Api/Models/SavedJoke.cs
```diff
@@ -7,5 +7,9 @@ namespace JokesOnYou.Web.Api.Models
...
+        public User User { get; set; }
```
chrisK00, line 13: Same as the other joke comment

Server/JokesOnYou.Web.Api/Models/Tag.cs
```diff
@@ -7,5 +7,11 @@ namespace JokesOnYou.Web.Api.Models
 {
     public class Tag
     {
+        public int Id { get; set; }
+        public string Name { get; set; }
+        public DateTime Created { get; set; }
```
chrisK00, line 12: Just pointing this out so we dont miss a class

Server/JokesOnYou.Web.Api/Models/Tag.cs
```diff
@@ -7,5 +7,11 @@ namespace JokesOnYou.Web.Api.Models
...
+        public List<Joke> Jokes { get; set; }
+        public User Owner { get; set; }
```
chrisK00, line 14: another one

Server/JokesOnYou.Web.Api/Profiles/JokesProfile.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Threading.Tasks;
+
+namespace JokesOnYou.Web.Api.Profiles
+{
+    public class JokesProfile : Profile
+    {
+        public JokesProfile()
+        {
+            CreateMap<Joke, JokeReplyDto>();
+            CreateMap<User, JokeAuthorDto>();
```
chrisK00, line 16: i would argue the user map should belong to userprofile

Server/JokesOnYou.Web.Api/Repositories/Interfaces/IJokesRepository.cs
```diff
@@ -7,5 +9,9 @@ namespace JokesOnYou.Web.Api.Repositories.Interfaces
 {
     public interface IJokesRepository
     {
+        public Task CreateJokeAsync(Joke joke);
+        public Task<IEnumerable<Joke>> GetAllJokesAsync();
+        public Task<IEnumerable<Joke>> GetJokesByPremiseAsync(string premise);
```
chrisK00, line 14: - Is the "public" keyword automatic when your doing interfaces or are you typing it? considering all members in an interface must be public
- GetJokesByPremise - bypremise is a filtering, so if we want multiple filtering that means we need a method for every single one. A better approach would be to use the GetAllJokesAsync() and in that method you send down a class filled with all possible filters

Server/JokesOnYou.Web.Api/Repositories/JokesRepository.cs
```diff
@@ -8,5 +14,43 @@ namespace JokesOnYou.Web.Api.Repositories
...
+        private readonly DataContext _context;
+
+        public JokesRepository(DataContext context)
+        {
+            _context = context;
+        }
+
+        public async Task CreateJokeAsync(Joke joke)
+        {
+            await _context.Jokes.AddAsync(joke);
```
chrisK00, line 26: We dont need to await one liners since nothing is happening afterwards. The addasync sends back a value task thought so you could add the ".AsTask()" at the end of the method

Server/JokesOnYou.Web.Api/Repositories/JokesRepository.cs
```diff
@@ -8,5 +14,43 @@ namespace JokesOnYou.Web.Api.Repositories
...
+        }
+
+        public async Task<IEnumerable<Joke>> GetJokesByPremiseAsync(string premise)
+        {
+            var lowerPremise = premise.ToLower();
+            var jokes = await _context.Jokes.Where(joke => joke.Premise.ToLower() == lowerPremise).ToListAsync();
```
chrisK00, line 32: maybe we could have like a normalized name like identity framework does

Server/JokesOnYou.Web.Api/Repositories/JokesRepository.cs
```diff
@@ -8,5 +14,43 @@ namespace JokesOnYou.Web.Api.Repositories
...
+
+            return jokes;
+        }
+
+        public async Task<IEnumerable<Joke>> GetAllJokesAsync()
+        {
+            var jokes = await _context.Jokes.Include( joke => joke.Author).ToListAsync();
```
chrisK00, line 39: We can return this directly without awaiting

Server/JokesOnYou.Web.Api/Repositories/JokesRepository.cs
```diff
@@ -8,5 +14,43 @@ namespace JokesOnYou.Web.Api.Repositories
...
+            return jokes;
+        }
+
+        public async Task<IEnumerable<JokeReplyDto>> GetAllJokeDtosAsync()
+        {
+            var configuration = new MapperConfiguration(cfg =>
+            {
+                cfg.CreateMap<Joke, JokeReplyDto>();
```
chrisK00, line 47: Should be in a profile and if im not mistaken youve added them already

Server/JokesOnYou.Web.Api/Services/Interfaces/IJokesService.cs
```diff
@@ -7,5 +8,7 @@ namespace JokesOnYou.Web.Api.Services.Interfaces
 {
     public interface IJokesService
     {
+        public Task<JokeReplyDto> CreateJokeAsync(JokeCreateDto jokeCreateDto, string userId);
```
chrisK00, line 11: tip: have the userId on the dto (it doesnt matter if someone tries to make a normal api request and sees oo i can enter user id because we will overwrite it with the userid from token)

Server/JokesOnYou.Web.Api/Services/JokesService.cs
```diff
@@ -10,10 +14,68 @@ namespace JokesOnYou.Web.Api.Services
...
+        public JokesService(IJokesRepository jokesRepo, IUserRepository userRepository, IUnitOfWork unitOfWork, IMapper mapper)
         {
             _jokesRepo = jokesRepo;
+            _mapper = mapper;
+            _userRepository = userRepository;
+            _unitOfWork = unitOfWork;
+        }
+        public async Task<JokeReplyDto> CreateJokeAsync(JokeCreateDto jokeCreateDto, string userId)
+        {
+            jokeCreateDto.Premise = jokeCreateDto.Premise.Trim();
```
chrisK00, line 30: On the dto entity you can make use of the "set" and trim there. Im not sure if its a bad practice to trim tho considering this is the views job. Might wanna ask someone @PoisnFang

Server/JokesOnYou.Web.Api/Services/JokesService.cs
```diff
@@ -10,10 +14,68 @@ namespace JokesOnYou.Web.Api.Services
...
+            }
+
+            var jokeReplyDto = _mapper.Map<JokeReplyDto>(joke);
+
+            return jokeReplyDto;
+        }
+
+        private async Task<bool> IsJokeDuplicate(JokeCreateDto jokeCreateDto)
+        {
+            var jokes = await _jokesRepo.GetJokesByPremiseAsync(jokeCreateDto.Premise);
```
chrisK00, line 58: Might wanna refactor this a bit. This is something that belongs to the repository. In there i would also just pass down the  punchline since thats all your checking, and i wouldnt wanna load the items to memory. So you can just do something like

return _context.jokes.any(x => x.punchline == punchline)

chrisK00 (changes requested):
Hmu on discord and id love to explain/help with anything that is confusing or unclear

Server/JokesOnYou.Web.Api/Services/JokesService.cs
chrisK00: Yea its fine. Didnt see this until now github.com likes to hide things lol

Server/JokesOnYou.Web.Api/Services/JokesService.cs
chrisK00: what other method could have saved this? Yeah service is the one handling the saving

Server/JokesOnYou.Web.Api/Services/JokesService.cs
chrisK00: Whats the reason for returning a joke here considering we are creating it? So if the other joke doesnt reach the db then we never get here since we make sure the stuff were saved.  Im not sure what kinda edge cases there are here @PoisnFang

HellHunterMax: The reason i would return the joke here is to show it to the person that created it. just like you would show a Registered user its account after registering. but it could be done differently of course.

Server/JokesOnYou.Web.Api/Repositories/Interfaces/IJokesRepository.cs
HellHunterMax: - Yes I am typing them in, I know it does not matter. I like to explicitly type stuff so I don't make any mistakes of forgetting it. I like to get used to stuff, so wherever I can I'll type it.
- Yes its technically a "filter" Though I needed this to check if a joke already exists in the DB. so its not the whole filtering issue that can be done by someone else and then this could get integrated.

Server/JokesOnYou.Web.Api/Repositories/JokesRepository.cs
HellHunterMax: that's a good thing to add to the Joke would save a lot of computation time.

Server/JokesOnYou.Web.Api/Services/JokesService.cs
HellHunterMax: This is just so all jokes that go into the database are trimmed of whitespace so we have as low of a duplicate as possible. (will not eliminate it of course )

Server/JokesOnYou.Web.Api/Profiles/JokesProfile.cs
HellHunterMax: ill create an issue to move it once the User part has been merged with dev.

Server/JokesOnYou.Web.Api/Models/Reason.cs
HellHunterMax: Yes but i think adding they are concerns for later?
when we know exactly which ones to add.
could be a good separate issue.

Server/JokesOnYou.Web.Api/Services/JokesService.cs
HellHunterMax: I thought because its Async another method somewhere else could save the Database before this one gets called ?
but if that's not possible then no check is needed.

chrisK00: Ahh i get it, Yeah changes can be saved in other places. But isnt this check was to see that there were actually any changes saved and this save was successfull

Server/JokesOnYou.Web.Api/Services/JokesService.cs
chrisK00: I guess you can just return the joke if the things you wanna send back does not need to be recieved from the db (since you already have them)

Server/JokesOnYou.Web.Api/Models/Reason.cs
chrisK00: Sure ill make it

Server/JokesOnYou.Web.Api/Services/JokesService.cs
HellHunterMax: ill add a second check for when its not saved to see if its in the database ;)

PR #106, opened 2021-04-20
Implemented GetJoke feature

### LINKED ISSUE #72: Endpoint: Get Joke
Create the whole line for GetJoke API call.

- [ ] GetJoke inside controller HTTP get
- [ ] GetJoke inside JokeService
- [ ] GetJoke inside JokeRepository

Don't forget to create a Dto for the return.

chrisK00:
closes #72 

Theres no need to check if the item is null when just returning a get request to the client. If the response is empty the client recieves a NoContent.

Tested with theese urls: 
- https://localhost:5001/api/Jokes/1 (existing joke) => Ok(JokeDto)
- https://localhost:5001/api/Jokes/55 (doesnt exist) => NoContent()

### DIFF
--- a/Server/JokesOnYou.Web.Api/Controllers/JokesController.cs
+++ b/Server/JokesOnYou.Web.Api/Controllers/JokesController.cs
@@ -1,11 +1,9 @@
-﻿using JokesOnYou.Web.Api.DTOs;
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+using JokesOnYou.Web.Api.DTOs;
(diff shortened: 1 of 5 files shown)


HellHunterMax (approved):
works and looks good!
Maybe change the var joke into jokeDto in controller but its not necessary.

chrisK00:
> works and looks good!
> Maybe change the var joke into jokeDto in controller but its not necessary.

Personal opinion but i prefer to not name them dto unless we actually have to know its a dto, considering we are just returning. Thanks for the feedback!

stutwoby (approved):

PR #107, opened 2021-04-22
Add user service

JBSE:
Just need a quick review. Seems too simple to be correct haha. Should I be caching the results, should I be authenticating anything?

### DIFF
--- a/Server/JokesOnYou.Web.Api/Controllers/UserController.cs
+++ b/Server/JokesOnYou.Web.Api/Controllers/UserController.cs
@@ -33,6 +33,12 @@ namespace JokesOnYou.Web.Api.Controllers
             throw new NotImplementedException();
         }
 
+        [HttpGet("GetAllUsers")]
+        public async Task<ActionResult> GetAll()
+        {
+            return Ok(_userService.GetAll());
+        }
+
         [HttpPatch]
         public async Task<ActionResult> UpdateUser(UserUpdateDTO userUpdateDTO)
         {
--- a/Server/JokesOnYou.Web.Api/Services/UserService.cs
+++ b/Server/JokesOnYou.Web.Api/Services/UserService.cs
@@ -25,7 +25,7 @@ namespace JokesOnYou.Web.Api.Services
 
         public async Task<IEnumerable<User>> GetAll()
         {
-            throw new NotImplementedException();
+            return await _userRepository.GetUsersAsync();
         }
 
         public async Task<User> GetUserById(string id)


Server/JokesOnYou.Web.Api/Services/UserService.cs
```diff
@@ -25,7 +25,7 @@ public async Task DeleteUser(string id)
 
         public async Task<IEnumerable<User>> GetAll()
```
HellHunterMax, line 26: Change name to GetAllAsync since its an async task. 
Could even change to GetAllUsersAsync to be more specific.

Server/JokesOnYou.Web.Api/Controllers/UserController.cs
```diff
@@ -33,6 +33,12 @@ public async Task<ActionResult> GetUser(string id)
             throw new NotImplementedException();
         }
 
+        [HttpGet("GetAllUsers")]
+        public async Task<ActionResult> GetAll()
+        {
+            return Ok(_userService.GetAll());
```
HellHunterMax, line 39: Needs to be awaited since GetAll is Async.

Server/JokesOnYou.Web.Api/Services/UserService.cs
```diff
@@ -25,7 +25,7 @@ public async Task DeleteUser(string id)
...
         {
-            throw new NotImplementedException();
+            return await _userRepository.GetUsersAsync();
```
HellHunterMax, line 28: Returning a full user model would not be good.
Use a DTO (Data Transfer Object) and only return the information that you want to return.
We don't want the receiver of this list to get hashed passwords.

Server/JokesOnYou.Web.Api/Controllers/UserController.cs
```diff
@@ -33,6 +33,12 @@ public async Task<ActionResult> GetUser(string id)
             throw new NotImplementedException();
         }
 
+        [HttpGet("GetAllUsers")]
+        public async Task<ActionResult> GetAll()
```
HellHunterMax, line 37: Add what is being returned inside the ActionResult<Type here (DTO)>

HellHunterMax (changes requested):
some tips:
1. Try and use meaningful commit messages. [See this article](https://medium.com/walmartglobaltech/check-out-these-5-git-tips-before-your-next-commit-c1c7a5ae34d1)
2. When Opening a PR fill it in as accurately as possible. "initial commit" does not say anything about what the PR is for.
3. if you need some help but don't want a full PR you could always open a draft. and change it when its ready.
4. did you test your code ?
5. I personally like to cache the results but its not needed.
6. You Could Authenticate it yes but its not needed yet, it can be added later on when adding and managing roles comes in.

Your code looks clean and good, needs some changes to work properly though.
Please react to the comments I made and close the conversations when finished so we can see how you're doing and what still needs work.
Good work! 🎉

JBSE:
@HellHunterMax Ah, I didn't realise you could create drafts. 

At work we typically raise PR's then abandon if needed. I just wanted this reviewed, I never intended this to get merged. Thanks!

PR #109, opened 2021-04-23
Feature/71 create joke

### LINKED ISSUE #71: Endpoint: Create Joke

HellHunterMax:
Made this new PR because the old one seemed bugged.

Fixed all the issues in the last PR #105 

# This implements endpoint Create a joke.
closes #71.

### how it works:
1. Controller takes the call.
2. controller gets the User Id from the claim.
3. controller gives the JokeDto and User Id to the Service.
4. Service Gets the User from the UserRepo.
5. Service maps the JokeDto into a Joke.
6. service places the User into Joke.Author.
7. Service gives the joke to the JokesRepo.
8. JokesRepo places the Joke into DB.
9. Service asks UnitOfWork to save the DB.
10. Service Places the Joke into a JokeReplyDto
11. service returns the JokeReplyDto to the Controller
12. controller Returns the JokeReplyDto to the caller.

### issues I had on this:
1. getting the user Id without giving it though the model. we could mix both for extra safety?
2. how to save the Database.
3. Do i call the UserRepo inside the JokesService or should JokesService call the UserService?
4. should we call the database after saving and check if its in there ?

### how They were fixed:
1. Chris had a nice extension for this.
2. Saving is done through UnitOfWork.
3. Calling the UserRepo is fine.
4. ???


Server/JokesOnYou.Web.Api/Controllers/JokesController.cs
```diff
@@ -21,6 +22,16 @@ public JokesController(IJokesService jokesService)
             _jokesService = jokesService;
         }
 
+        [Authorize(Roles = "Registered,Admin")]
```
chrisK00, line 23: idea: maybe member instead of registered?

Server/JokesOnYou.Web.Api/Controllers/JokesController.cs
```diff
@@ -21,6 +22,16 @@ public JokesController(IJokesService jokesService)
...
+        [HttpPost]
+        public async Task<ActionResult<JokeReplyDto>> CreateJokeAsync(JokeCreateDto jokeCreateDto)
+        {
+
+            jokeCreateDto.UserId = ClaimsPrincipalExtension.GetUserId(User);
+            var jokeReplyDto = await _jokesService.CreateJokeAsync(jokeCreateDto);
+
+            return Ok(jokeReplyDto);
```
chrisK00: no need to wrap this in a Ok() in order to get type safety

Server/JokesOnYou.Web.Api/DTOs/JokeCreateDto.cs
```diff
@@ -7,5 +7,10 @@ namespace JokesOnYou.Web.Api.DTOs
 {
     public class JokeCreateDto
     {
+        public string Premise { get; set; }
+        public string NormalizedPremise { get; set; }
```
chrisK00, line 11: Whats the purpose of normalizedpremise on the dto?  Shouldnt this just be on the Joke entity and make the premise required then inside its setter you also set the normalizedpremise but value.toupper?

Server/JokesOnYou.Web.Api/Migrations/20210423022205_AddNormalizedJokes.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    public partial class AddNormalizedJokes : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<string>(
+                name: "NormalizedPremise",
+                table: "Jokes",
+                type: "TEXT",
+                nullable: true);
```
chrisK00, line 13: I dont think this should be nullable

Server/JokesOnYou.Web.Api/Migrations/20210423023914_ChangedModelIntoId.cs
```diff
@@ -0,0 +1,34 @@
...
+    public partial class ChangedModelIntoId : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.DropForeignKey(
+                name: "FK_Jokes_AspNetUsers_AuthorId",
+                table: "Jokes");
+
+            migrationBuilder.DropIndex(
+                name: "IX_Jokes_AuthorId",
```
chrisK00: Did you configure the FK in modelbuilder?

Server/JokesOnYou.Web.Api/Models/LikedTags.cs
```diff
@@ -8,9 +8,9 @@ namespace JokesOnYou.Web.Api.Models
     public class LikedTags
     {
         public int Id { get; set; }
-        public Joke Joke { get; set; }
-        public DateTime SavedDate { get; set; }
-        public List<Tag> Tags { get; set; }
-        public User User { get; set; }
+        public int JokeId { get; set; }
+        public DateTime SavedDate { get; set; } = DateTime.Now;
+        public List<int> Tags { get; set; }
```
chrisK00: always initialize lists. Would also recommend using ICollection

Server/JokesOnYou.Web.Api/Repositories/Interfaces/IJokesRepository.cs
```diff
@@ -9,7 +9,10 @@ namespace JokesOnYou.Web.Api.Repositories.Interfaces
 {
     public interface IJokesRepository
     {
-        public Task<IEnumerable<Joke>> GetAllJokesAsync();
-        public Task<IEnumerable<JokeReplyDto>> GetAllJokeDtosAsync();
+        public Task CreateJokeAsync(Joke joke);
+        public Task<List<Joke>> GetAllJokesAsync();
```
chrisK00: Consider changing from list to ienumerable

Server/JokesOnYou.Web.Api/Repositories/Interfaces/IJokesRepository.cs
```diff
@@ -9,7 +9,10 @@ namespace JokesOnYou.Web.Api.Repositories.Interfaces
...
+        public Task<bool> DoesJokeExist(JokeCreateDto jokeCreateDto);
```
chrisK00: cant we check if a joke exists without having to pass down a jokecreatedto?

Server/JokesOnYou.Web.Api/Services/JokesService.cs
```diff
@@ -12,12 +14,58 @@ namespace JokesOnYou.Web.Api.Services
...
             _mapper = mapper;
+            _userRepository = userRepository;
+            _unitOfWork = unitOfWork;
+        }
+
+        public async Task<JokeReplyDto> CreateJokeAsync(JokeCreateDto jokeCreateDto)
+        {
+            TrimAndNormalizeJokeCreateDto(jokeCreateDto);
+
+            var isDuplicate = await _jokesRepo.DoesJokeExist(jokeCreateDto);
```
chrisK00: you can insert the await.... inside the if statement since it returns a boolean

Server/JokesOnYou.Web.Api/Services/JokesService.cs
```diff
@@ -12,12 +14,58 @@ namespace JokesOnYou.Web.Api.Services
...
+            }
+
+            var joke = _mapper.Map<Joke>(jokeCreateDto);
+            joke.AuthorId = user.Id;
+            await _jokesRepo.CreateJokeAsync(joke);
+
+            var saved = await _unitOfWork.SaveAsync();
+            if (!saved)
+            {
+                if(await _jokesRepo.GetJokeDtoAsync(joke.Id) == null)
```
chrisK00, line 49: Are you trying to get the joke after the fail saved?

chrisK00 (changes requested):

Server/JokesOnYou.Web.Api/Controllers/JokesController.cs
HellHunterMax: Yea thats fine, this is just placeholder anyway ;)

Server/JokesOnYou.Web.Api/Migrations/20210423023914_ChangedModelIntoId.cs
HellHunterMax: no because its just a string and not a User anymore its not automatically linked to a FK ill try to change that.

Server/JokesOnYou.Web.Api/Repositories/Interfaces/IJokesRepository.cs
HellHunterMax: how ? you have to know what to look for to be able to know if it exists?

Server/JokesOnYou.Web.Api/Services/JokesService.cs
HellHunterMax: Yes its possible, I thought this might be more readable but I guess it could be readable enough.

Server/JokesOnYou.Web.Api/Services/JokesService.cs
HellHunterMax: yea i find this whole part weird. I have to look into Unit of work more i think.
What would your advice be how to handle this ?

Server/JokesOnYou.Web.Api/Repositories/Interfaces/IJokesRepository.cs
chrisK00: Looked at your method so all you would need is a normalized premise and punchline to be sent down, making this method reusable not limited to a specific dto

Server/JokesOnYou.Web.Api/Services/JokesService.cs
chrisK00: The point of the unitofwork is to save multiple changes at once so if one fails they all fail in order to avoid inconsistent data in our DB.

How id handle it is almost the same as you.
1. make an if trying to save changes: If (!await _unitOfWork.SaveAsync())
                                throw new appexception
2. Now we can map the joke to a jokereplydto (Yes i now remembered we dont need to get the item from the db!) Since i noticed you make a joke here before passing down, that joke is a reference type so it will also be updated here with properties such as the id so this is the same as the item in db because EF is tracking it

chrisK00 (changes requested):
Almost there!

Server/JokesOnYou.Web.Api/Repositories/Interfaces/IJokesRepository.cs
HellHunterMax: duh :P

Server/JokesOnYou.Web.Api/Services/JokesService.cs
HellHunterMax: but if another part of the app saves the database in the mean time this will fail ?
while it actually succeeded. is that possible ?

Server/JokesOnYou.Web.Api/Migrations/20210423022205_AddNormalizedJokes.cs
HellHunterMax: please check if i made good changes.

Server/JokesOnYou.Web.Api/Migrations/20210423023914_ChangedModelIntoId.cs
HellHunterMax: did i do it correct?

stutwoby (approved):

chrisK00 (approved):
Looking good!

PR #111, opened 2021-04-26
Feature 74 update joke

### LINKED ISSUE #74: Endpoint: Update Joke
Create the whole line for Update Joke API call.

- [ ] UpdateJoke inside controller HTTP put
- [ ] UpdateJoke inside JokeService
- [ ] UpdateJoke inside JokeRepository

Use AutoMapper if you can.

ofu997:
this works but it might be organized better.


Server/JokesOnYou.Web.Api/Controllers/JokesController.cs
```diff
@@ -29,5 +31,26 @@ public async Task<ActionResult<IEnumerable<JokeReplyDto>>> GetAllJokesAsync()
...
 
+        [AllowAnonymous]
+        [HttpGet("{id}")]
+        public async Task<JokeReplyDto> GetJokeById(int id)
+        {
+            var jokeReplyDto = await _jokesService.GetJokeReplyById(id);
+            return jokeReplyDto;
+        }
+
+        // Create the whole line for Update Joke API call.
```
chrisK00: what are theese comments?

Server/JokesOnYou.Web.Api/Controllers/JokesController.cs
```diff
@@ -29,5 +31,26 @@ public async Task<ActionResult<IEnumerable<JokeReplyDto>>> GetAllJokesAsync()
...
+        // UpdateJoke inside JokeService
+        // UpdateJoke inside JokeRepository
+        // Use AutoMapper if you can.
+
+        [HttpPatch]
+        [AllowAnonymous]
+        public async Task<ActionResult<JokeUpdateDTO>> UpdateJoke(JokeUpdateDTO jokeUpdateDTO)
+        {
+            await _jokesService.UpdateJoke(jokeUpdateDTO);
+            return Ok();
```
chrisK00: Use Nocontent() instead, it means the request was successfull but we have nothing to send back

Server/JokesOnYou.Web.Api/Controllers/JokesController.cs
```diff
@@ -29,5 +31,26 @@ public async Task<ActionResult<IEnumerable<JokeReplyDto>>> GetAllJokesAsync()
...
+        }
+
+        // Create the whole line for Update Joke API call.
+        // UpdateJoke inside controller HTTP put
+        // UpdateJoke inside JokeService
+        // UpdateJoke inside JokeRepository
+        // Use AutoMapper if you can.
+
+        [HttpPatch]
+        [AllowAnonymous]
```
chrisK00, line 55: This is a http put, patch is when your updating a single property and would require json patch. Should also be authorized

Server/JokesOnYou.Web.Api/DTOs/JokeUpdateDTO.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace JokesOnYou.Web.Api.DTOs
+{
+    // Does not update joke's id, likes, author, uploaddate, or dislikes. Just Premise and Punchline
```
chrisK00: You can use xml comments here three backslashes ///

Server/JokesOnYou.Web.Api/Profiles/JokesProfile.cs
```diff
@@ -14,6 +14,7 @@ public JokesProfile()
         {
             CreateMap<Joke, JokeReplyDto>();
             CreateMap<User, JokeAuthorDto>();
+            CreateMap<Joke, JokeUpdateDTO>();
```
chrisK00: We map From a incoming jokeupdatedto To a Joke so should be reversed

Server/JokesOnYou.Web.Api/Repositories/Interfaces/IJokesRepository.cs
```diff
@@ -11,5 +11,9 @@ public interface IJokesRepository
     {
         public Task<IEnumerable<Joke>> GetAllJokesAsync();
         public Task<IEnumerable<JokeReplyDto>> GetAllJokeDtosAsync();
+        public Task<Joke> GetJokeToUpdate(JokeUpdateDTO jokeUpdateDTO);
+        public Task<JokeReplyDto> GetJokeReplyById(int id);
+        public Task<Joke> GetJokeById(int id); 
+        public Task Save();
```
chrisK00: IUnitOfWork takes care of saving

Server/JokesOnYou.Web.Api/Repositories/Interfaces/IJokesRepository.cs
```diff
@@ -11,5 +11,9 @@ public interface IJokesRepository
     {
         public Task<IEnumerable<Joke>> GetAllJokesAsync();
         public Task<IEnumerable<JokeReplyDto>> GetAllJokeDtosAsync();
+        public Task<Joke> GetJokeToUpdate(JokeUpdateDTO jokeUpdateDTO);
```
chrisK00: Use getjokebyid instead

Server/JokesOnYou.Web.Api/Repositories/JokesRepository.cs
```diff
@@ -35,5 +35,40 @@ public async Task<IEnumerable<JokeReplyDto>> GetAllJokeDtosAsync()
...
+        public async Task<JokeReplyDto> GetJokeReplyById(int id)
+        {
+            var joke = await GetJokeById(id); 
+
+            if (joke == null)
+            {
+                throw new NotImplementedException("can't find joke"); 
+            }
+
+            var jokeReplyDto = _mapper.Map<Joke, JokeReplyDto>(joke);
```
chrisK00: _mapper.Map<Joke, JokeReplyDto> its enough to specify which entity we are mapping to which is JokeReplyDto

Server/JokesOnYou.Web.Api/Services/JokesService.cs
```diff
@@ -26,5 +26,29 @@ public async Task<IEnumerable<JokeReplyDto>> GetAllJokeDtosAsync()
...
+            // get the joke we need to update
+            // set fields with jokeUpdateDTO
+            var jokeToUpdate = await _jokesRepo.GetJokeToUpdate(jokeUpdateDTO);
+
+            if (jokeToUpdate == null)
+            {
+                throw new NotImplementedException("can't find joke to update");
+            }
+
+            jokeToUpdate.Premise = jokeUpdateDTO.Premise;
```
chrisK00: We can use automapper here. So instead of mapping to a new item we can map to an existing

it would look like this map(From, To)

Server/JokesOnYou.Web.Api/Services/JokesService.cs
```diff
@@ -26,5 +26,29 @@ public async Task<IEnumerable<JokeReplyDto>> GetAllJokeDtosAsync()
...
+            jokeToUpdate.Punchline = jokeUpdateDTO.Punchline;
+
+            await _jokesRepo.Save(); 
+        }
+
+        public async Task<JokeReplyDto> GetJokeReplyById(int id)
+        {
+            var jokeDtos = await _jokesRepo.GetJokeReplyById(id);
```
chrisK00: JokeDto*

chrisK00 (changes requested):
You got the idea. Lets make the code cleaner

Server/JokesOnYou.Web.Api/Services/JokesService.cs
```diff
@@ -26,5 +26,29 @@ public async Task<IEnumerable<JokeReplyDto>> GetAllJokeDtosAsync()
...
+
+        public async Task UpdateJoke(JokeUpdateDTO jokeUpdateDTO)
+        {
+            // get the joke we need to update
+            // set fields with jokeUpdateDTO
+            var jokeToUpdate = await _jokesRepo.GetJokeToUpdate(jokeUpdateDTO);
+
+            if (jokeToUpdate == null)
+            {
+                throw new NotImplementedException("can't find joke to update");
```
heijmn: Instead of throwing this exception here I think you should return a NotFound() result in the controller.

Server/JokesOnYou.Web.Api/Repositories/JokesRepository.cs
```diff
@@ -35,5 +35,40 @@ public async Task<IEnumerable<JokeReplyDto>> GetAllJokeDtosAsync()
...
+            await _context.SaveChangesAsync();
+        }
+
+        public async Task<JokeReplyDto> GetJokeReplyById(int id)
+        {
+            var joke = await GetJokeById(id); 
+
+            if (joke == null)
+            {
+                throw new NotImplementedException("can't find joke");
```
heijmn: NotImplemented seems to be the wrong exception in my opinion.

Server/JokesOnYou.Web.Api/Repositories/JokesRepository.cs
```diff
@@ -35,5 +35,40 @@ public async Task<IEnumerable<JokeReplyDto>> GetAllJokeDtosAsync()
...
+                throw new NotImplementedException("can't find joke"); 
+            }
+
+            var jokeReplyDto = _mapper.Map<Joke, JokeReplyDto>(joke);
+            return jokeReplyDto;
+        }
+
+        public async Task<Joke> GetJokeById(int id)
+        {
+            var joke = await _context.Jokes.FirstOrDefaultAsync(x => x.Id == id);
```
heijmn: @chrisK00, I saw this in other places as well. But shouldn't it be SingleOrDefault? Because we only expect the ID to exist once...?

Server/JokesOnYou.Web.Api/DTOs/JokeUpdateDTO.cs
```diff
@@ -0,0 +1,15 @@
...
+    public class JokeUpdateDTO
```
heijmn: What is the naming convention? I saw some classes named ...DTO and some ...Dto

HellHunterMax: Yea my bad.... I started with all capital like these and then remembered:
`For abreviations, use Pascal Case as well. For example Bmi` [Found in wiki](https://github.com/Code2Gether-Discord/JokesOnYou/wiki/Coding-Standards).

Server/JokesOnYou.Web.Api/Repositories/JokesRepository.cs
HellHunterMax: `SingleOrDefault` will go over the whole table no matter if its already found the object. it will continue checking if there is another object that satisfies the Predicate. If for whatever reason two objects are found it will return null.
`FirstOrDefault` will stop when 1 is found.

if I did not get this correct please let me know.

Server/JokesOnYou.Web.Api/Services/JokesService.cs
HellHunterMax: use `KeyNotFoundException` then the GlobalExeptionhandler will handle returning it.
 [GlobalExeptionHandler](https://github.com/Code2Gether-Discord/JokesOnYou/blob/dev/Server/JokesOnYou.Web.Api/Middlewares/ExceptionMiddleware.cs)

Server/JokesOnYou.Web.Api/Repositories/JokesRepository.cs
heijmn: SingleOrDefault will return an Exception if more than one object is found, not null. It will only return null when nothing is found. 

I prefer this method over FirstOrDefault because it ensures that only one item is found and otherwise let's you know via a Exception that something is wrong. However, FirstOrDefault will indeed be faster. While both are still N(1) I think.

I guess this is comes down to taste, I personally like the explicitly of saying: I want this one thing and I expect that there is exactly one thing that matches my search.

chrisK00: I use singleordefault when im searching by name for example because there is a chance we have duplicates since its us developers who did it, but when searching for ids i use firstordefault because our db should auto increment and not allow duplicates

Server/JokesOnYou.Web.Api/Services/JokesService.cs
ofu997: PR #106 , Feature #72 takes care of this

Server/JokesOnYou.Web.Api/Repositories/JokesRepository.cs
```diff
@@ -34,5 +35,11 @@ public Task<JokeReplyDto> GetJokeDtoAsync(int id)
             return _context.Jokes.ProjectTo<JokeReplyDto>(_mapper.ConfigurationProvider)
                 .FirstOrDefaultAsync(j => j.Id == id);
         }
+
+        public async Task<Joke> GetJokeById(int id)
+        {
+            var joke = await _context.Jokes.FirstOrDefaultAsync(x => x.Id == id);
```
chrisK00: We can return the joke directly here. Subjective thought but its good practice to follow the same pattern as the other methods, and we do save some memory.

Server/JokesOnYou.Web.Api/Services/Interfaces/IJokesService.cs
```diff
@@ -6,6 +6,7 @@ namespace JokesOnYou.Web.Api.Services.Interfaces
 {
     public interface IJokesService
     {
+        Task UpdateJoke(JokeUpdateDto jokeUpdateDto);
```
chrisK00: I think its a good idea to return the updated joke, just like we do when we create it

chrisK00 (changes requested):
Very well done Oliver! Clean code

ofu997 (commented):
reviewed

chrisK00 (approved):
Lets get this merged

HellHunterMax (approved):
great job! looks great like this really clean.
We would need to add some kind of safeguard so not just anyone could update the joke but this is good for now!

PR #112, opened 2021-04-26
Feature/110 severity range

### LINKED ISSUE #110: Range for Severity property and others
The reason class has a int property named Severity. So we probably need some sort fo acceptable range which can be achieved with build in data annotation Range. This also applies to any other properties that have some soer of rating or range. 

The question is what range do we use, 1-5?

chrisK00:
closes #110 
Severity can only be between the range 1-5

### DIFF
--- a/Server/JokesOnYou.Web.Api/DTOs/ReasonCreateDto.cs
+++ b/Server/JokesOnYou.Web.Api/DTOs/ReasonCreateDto.cs
@@ -1,11 +1,14 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Threading.Tasks;
+using System.ComponentModel.DataAnnotations;
 
 namespace JokesOnYou.Web.Api.DTOs
 {
     public class ReasonCreateDto
     {
+        [Required]
+        public string Name { get; set; }
+
+        [Required, Range(1, 5)]
+        public int Severity { get; set; }
     }
 }
--- a/Server/JokesOnYou.Web.Api/DTOs/ReasonCreateDto.cs
+++ b/Server/JokesOnYou.Web.Api/DTOs/ReasonCreateDto.cs
@@ -8,7 +8,7 @@ namespace JokesOnYou.Web.Api.DTOs
         [Required]
         public string Name { get; set; }
 
-        [Required, Range(1, 5)]
+        [Required, Range(1, 50)]
         public int Severity { get; set; }
     }
 }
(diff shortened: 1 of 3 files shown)


ofu997 (approved):
looks good

ofu997 (commented):
maybe a severity property belongs to flag. Like in football you can get a yellow or red flag based on severity.

chrisK00:
> maybe a severity property belongs to flag. Like in football you can get a yellow or red flag based on severity.

Yup the flag does have a Reason property. I noticed we havent migrated the new entities to the db tho

ofu997:
> > maybe a severity property belongs to flag. Like in football you can get a yellow or red flag based on severity.
> 
> Yup the flag does have a Reason property. I noticed we havent migrated the new entities to the db tho

i mean severity should belong to Flag, instead of to Reason.  So a message would say "you've been flagged, reason is "  ", this is a 3 on the severity scale"

HellHunterMax:
I'm of the opinion we should have a larger range to account for better customization.
Like:
you can have a maximum of 300 strikes before a ban. you could give someone 300 strikes for writing a full joke with only curse words.
but if someone uploads an empty toke you could give 10 strikes.
for 1 curse word in a joke you give 50 points.
if you only have 1-5 you don't have much room.
5 strikes == ban?
1 strike for empty joke ?
5 times you make a mistake for wherever reason. you get banned.
so but we could always change this in a discussion. the code is good.

HellHunterMax (approved):

chrisK00:
> > > maybe a severity property belongs to flag. Like in football you can get a yellow or red flag based on severity.
> > 
> > 
> > Yup the flag does have a Reason property. I noticed we havent migrated the new entities to the db tho
> 
> i mean severity should belong to Flag, instead of to Reason. So a message would say "you've been flagged, reason is " ", this is a 3 on the severity scale"

I think its because reason is supposed to be like more details in order to split up the entities, making it reusable. Think its just the name "reason" that is a bit unclear. @HellHunterMax

chrisK00:
> I'm of the opinion we should have a larger range to account for better customization.
> Like:
> you can have a maximum of 300 strikes before a ban. you could give someone 300 strikes for writing a full joke with only curse words.
> but if someone uploads an empty toke you could give 10 strikes.
> for 1 curse word in a joke you give 50 points.
> if you only have 1-5 you don't have much room.
> 5 strikes == ban?
> 1 strike for empty joke ?
> 5 times you make a mistake for wherever reason. you get banned.
> so but we could always change this in a discussion. the code is good.

Is there a specific reason for a such a high range? F.e would maybe 1-50 be enough?
10p empty joke (this is really bad for our site because users get tired from opening theese)
20p If its not a NSFW joke and contains curse word (so user needs 3x of this)

HellHunterMax:
Yea 1-50 should be enough

PR #113, opened 2021-04-26
Feature/75 delete joke

### LINKED ISSUE #75: Endpoint: Delete Joke
Create the whole line for Delete Joke API call.

- [ ] DeleteJoke inside controller HTTP Delete
- [ ] DeleteJoke  inside JokeService
- [ ] DeleteJoke inside JokeRepository

Niisc:
Added stuff necessary to delete jokes


Server/JokesOnYou.Web.Api/Controllers/JokesController.cs
```diff
@@ -29,5 +29,12 @@ public async Task<ActionResult<IEnumerable<JokeReplyDto>>> GetAllJokesAsync()
             return Ok(jokeDtos);
         }
 
+        [HttpDelete("{id}")]
+        public async Task<ActionResult<IEnumerable<JokeReplyDto>>> DeleteJoke(int id)
```
chrisK00: We are not returning anything rather than NoContent()

Server/JokesOnYou.Web.Api/Repositories/JokesRepository.cs
```diff
@@ -35,5 +48,15 @@ public async Task<IEnumerable<JokeReplyDto>> GetAllJokeDtosAsync()
 
             return jokeDtos;
         }
+
+        public void DeleteJoke(Joke joke)
+        {
+            _context.Jokes.Remove(joke);
+        }
+
+        public async Task<Joke> GetJokeByIdAsync(int id)
```
chrisK00, line 54: Does not need to be awaited

chrisK00 (changes requested):
Gj! Make sure to check the github actions warnings aswell

Server/JokesOnYou.Web.Api/Controllers/JokesController.cs
Niisc: Forgot to remove it, whoops

Server/JokesOnYou.Web.Api/Repositories/JokesRepository.cs
Niisc: ?

Niisc: What shouldn't be awaited?

chrisK00: return await _context.Jokes.FirstOrDefaultAsync(x => x.Id == id);
since its a one liner the caller can await it

chrisK00 (changes requested):

Server/JokesOnYou.Web.Api/Repositories/JokesRepository.cs
hartmannr76: It's actually generally suggested to await a task instead of directly returning it
[Reference](https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#prefer-asyncawait-over-directly-returning-task)

chrisK00: Its actually recommended by many senior devs. Since theres nothing after the code theres no point in waiting for it to finish inside the method, so for simple oneliners it is ok to do this. Yes it does help performance a little but also can be a problem with debugging but thats when you have bigger methods tho https://www.youtube.com/watch?v=Q2zDatDVnO0&t=458s

hartmannr76: I'm confused - Your comment suggests to remove the await and just return the task; but the video suggests await the task and he also uses the same reference to Fowler's guidance

hartmannr76: I'm confused - Your original comment suggests removing the await, but the video suggests leaving it and even references to the same async guidance I had linked. Is removing it more of a personal/project preference?

chrisK00: Yes i consider it a personal preference, f.e @N1c0z prefers performance so removing await would help with that. I cannot see the harm in removing await from one liners that are plain simple like this. if the method had more lines or was more complex it would make sense to use await, getting a joke from db with its id will always look the same so you wouldnt even debug this you would rather go to the parent

Niisc: I am confusion, remove or keep? Performance doesn't really matter since compiler cleans it all up.

Niisc: I mean, it _does_ matter,

hartmannr76: The perf impact is almost negligible. Heres a sample micro-benchmark I ran if you're curious about the results
```
// Setup
 public class TaskTester {
        private Task EmtpyTask() => Task.Delay(1);
        private async Task WaitedEmptyTask() => await Task.Delay(1);

        [Benchmark]
        public async Task AwaitParent() => await EmtpyTask();
        [Benchmark]
        public async Task AwaitChild() => await WaitedEmptyTask();
    }
```

```
BenchmarkDotNet=v0.12.1, OS=linuxmint 20
AMD Athlon(tm) 64 FX-62 Dual Core Processor, 1 CPU, 2 logical and 2 physical cores
.NET Core SDK=6.0.100-preview.3.21202.5
  [Host]     : .NET Core 6.0.0 (CoreCLR 6.0.21.20104, CoreFX 6.0.21.20104), X64 RyuJIT
  DefaultJob : .NET Core 6.0.0 (CoreCLR 6.0.21.20104, CoreFX 6.0.21.20104), X64 RyuJIT


|      Method |     Mean |     Error |    StdDev |   Median |
|------------ |---------:|----------:|----------:|---------:|
| AwaitParent | 4.637 ms | 0.0915 ms | 0.0811 ms | 4.627 ms |
|  AwaitChild | 4.610 ms | 0.1198 ms | 0.3533 ms | 4.733 ms |
```

But if it's a personal preference choice. I generally defer to David Fowler with many of my opinions, which is why my personal is to leave it; but that's not my call to make

chrisK00: Yup its def a very tiny difference. Well then as long as we are all on the same page i also suggest refactoring all the other places we are not using async/await together.

Niisc: So we just keep it like this?

HellHunterMax: > So we just keep it like this?

Yes if I can puzzle out these comments we keep it and change all others to make our code uniform!
Reason = better debugging.

**Always use async/await.**

chrisK00 (approved):

Server/JokesOnYou.Web.Api/Repositories/JokesRepository.cs
Niisc: So i'll leave it as is

HellHunterMax (approved):
There are some double using statements inside JokesRepository please remove them before merging.

PR #114, opened 2021-04-28
87 get all tags

### LINKED ISSUE #87: Endpoint: Get All Tags
Create the whole line for Get All Tags API call.

- [ ] GetTag inside controller HTTP get
- [ ] GetTag inside Service
- [ ] GetTag inside Repository

<details>
  <summary>How To / tips:</summary>

- Return an IEnumerable<TagReplyDto>
</details>

nabeelfarooqui98:
Added endpoint for get all tags and relevant methods in service and repository. Registered for dependency injection. Added sample data row in database.


Server/JokesOnYou.Web.Api/DTOs/TagReplyDto.cs
```diff
@@ -7,5 +7,11 @@ namespace JokesOnYou.Web.Api.DTOs
 {
     public class TagReplyDto
     {
+        public int Id { get; set; }
+        public string Name { get; set; }
+        public DateTime Created { get; set; }
+        public List<JokeReplyDto> Jokes { get; set; }
+        public UserReplyDTO Owner { get; set; }
```
heijmn: Won't this result in a security issue where the API provides to much data to the client (like token) / email address? Might be better to create a separate DTO for this? TagOwnerDTO with only id and username.

Server/JokesOnYou.Web.Api/DTOs/TagReplyDto.cs
```diff
@@ -7,5 +7,11 @@ namespace JokesOnYou.Web.Api.DTOs
 {
     public class TagReplyDto
     {
+        public int Id { get; set; }
+        public string Name { get; set; }
+        public DateTime Created { get; set; }
+        public List<JokeReplyDto> Jokes { get; set; }
```
heijmn: Not sure if this is a good thing to do... Won't this result in all tags being retrieved AND all jokes attached to those tags (I'm not sure)? Wouldn't it be better to have a dedicated call for that in the repo?

Server/JokesOnYou.Web.Api/Profiles/TagProfile.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace JokesOnYou.Web.Api.Profiles
+{
+    public class TagProfile : Profile
+    {
+        public TagProfile()
+        {
+            CreateMap<Tag, TagReplyDto>();
+            CreateMap<Joke, JokeReplyDto>();
+            CreateMap<User, UserReplyDTO>();
```
heijmn: Asking to learn here... These maps are already defined in JokesProfile. Won't they overlap with each other?

Server/JokesOnYou.Web.Api/DTOs/TagReplyDto.cs
nabeelfarooqui98: Yes that is what is happening at the moment. All tags with all jokes. I suppose that is a lot of unnecessary data. I actually saw the DTO and assumed that is what was planned

Server/JokesOnYou.Web.Api/DTOs/TagReplyDto.cs
nabeelfarooqui98: I actually dont know the practices of DTOs and their naming convention and what is supposed to return what. I saw this Tag DTO already created and assumed it was intended

Server/JokesOnYou.Web.Api/Profiles/TagProfile.cs
nabeelfarooqui98: Yes i did see that. But wouldnt that make this endpoint dependent on that profile?

heijmn: Yeah, I don't know how to do this actually. Normally I would just put all mappings in one big file. But I have no idea if that is the way forward.

Server/JokesOnYou.Web.Api/Controllers/TagController.cs
```diff
@@ -15,5 +20,13 @@ public TagController(ITagService tagService)
         {
             _tagService = tagService;
         }
+
+        [AllowAnonymous]
+        [HttpGet]
+        public async Task<ActionResult<IEnumerable<TagReplyDto>>> GetAllJokesAsync()
```
chrisK00: Did you mean GetAllTags?

Server/JokesOnYou.Web.Api/DTOs/TagReplyDto.cs
chrisK00: > Yes that is what is happening at the moment. All tags with all jokes. I suppose that is a lot of unnecessary data. I actually saw the DTO and assumed that is what was planned

Lets say you wanna get all jokes in a category. Then you would pass down the category name because its basically a filter. So get tags i think would be just for managing tags or our client to know what tags are available so they can design the front end

Server/JokesOnYou.Web.Api/DTOs/TagReplyDto.cs
chrisK00: dto = data transfer object. its also known as ViewModel because its just to give the correct data back to our client

Server/JokesOnYou.Web.Api/Data/DataContext.cs
```diff
@@ -15,6 +15,7 @@ public DataContext(DbContextOptions<DataContext> options) : base(options)
         }
     
         public DbSet<Joke> Jokes { get; set; }
+        public DbSet<Tag> Tag { get; set; }
```
chrisK00: Pluralize the name

Server/JokesOnYou.Web.Api/Profiles/TagProfile.cs
chrisK00: To follow SRP we have a profile for every entity no need to register the same mapping twice. When you register automapper in DI i bet it uses reflection to find all profiles.

Server/JokesOnYou.Web.Api/Repositories/Interfaces/ITagRepository.cs
```diff
@@ -7,5 +8,6 @@ namespace JokesOnYou.Web.Api.Repositories.Interfaces
 {
     public interface ITagRepository
     {
+        public Task<IEnumerable<TagReplyDto>> GetAllTagDtosAsync();
```
chrisK00: All things inside a interface must be public so no need to explicitly type it

chrisK00 (changes requested):

Server/JokesOnYou.Web.Api/Services/Interfaces/ITagService.cs
```diff
@@ -7,5 +8,6 @@ namespace JokesOnYou.Web.Api.Services.Interfaces
 {
     public interface ITagService
     {
+        public Task<IEnumerable<TagReplyDto>> GetAllTagDtosAsync();
```
heijmn: See comment in ITagRepository. Does not need to be public.

Server/JokesOnYou.Web.Api/DTOs/TagReplyDto.cs
nabeelfarooqui98: So do I need to make a new DTO? Or just not fetch the jokes from db.

What about the Owner UserReplyDTO?

chrisK00: You can remove the Jokes property from this dto, and yeah dont get the jokes from the db since that will become a performance issue. 

If we wanna show the client who made this dto we could just have the username instead of the actual user

Server/JokesOnYou.Web.Api/Repositories/TagRepository.cs
```diff
@@ -8,5 +13,19 @@ namespace JokesOnYou.Web.Api.Repositories
...
+        public TagRepository(DataContext context, IMapper mapper)
+        {
+            _context = context;
+            _mapper = mapper;
+        }
+        public async Task<IEnumerable<TagReplyDto>> GetAllTagDtosAsync()
+        {
+            var tagDtos = await _context.Tags.ProjectTo<TagReplyDto>(_mapper.ConfigurationProvider).ToListAsync();
+
+            return tagDtos;
```
bschaatsbergen: we can immediately return on the method instead of explicitly allocating memory first

Server/JokesOnYou.Web.Api/Models/Tag.cs
```diff
@@ -10,6 +11,7 @@ public class Tag
         public int Id { get; set; }
         public string Name { get; set; }
         public DateTime Created { get; set;  } = DateTime.Now;
+        [NotMapped]
```
chrisK00, line 14: If this is notmapped, what is this list for since its data wont be persisted?

chrisK00 (approved):
Well done!

Server/JokesOnYou.Web.Api/Models/Tag.cs
nabeelfarooqui98: i honestly have no idea. this came in when I merged dev into this branch. Had to use the NotMapped attribute otherwise it wouldnt run. 

I'm not sure how table joins work in code models.

HellHunterMax (approved):
good tested and works!

PR #117, opened 2021-04-30
Delete tag #88

### LINKED ISSUE #88: Endpoint: Delete Tag
Create the whole line for Delete Tag API call.

- [x] DeleteTag inside controller HTTP Delete
- [x] DeleteTag inside Service
- [x] DeleteTag inside Repository

oguzhanaksoy:
DeleteTag endpoints are added.
Delete Tag, related service methods added
Delete Tag, related repository methods are added.
Tag added as DbSet member.

You can delete with single or multiple tags.
Acceptable parameters:
- TagCreateDto, 
- int, 
- List<TagCreateDto>, 
- int[]


Server/JokesOnYou.Web.Api/Controllers/TagController.cs
```diff
@@ -15,5 +20,41 @@ public TagController(ITagService tagService)
         {
             _tagService = tagService;
         }
+
+        [HttpDelete]
+        public async Task<ActionResult<bool>> DeleteTag(TagCreateDto tag)
```
stutwoby: Is this the appropriate DTO to delete a tag?

Server/JokesOnYou.Web.Api/Controllers/TagController.cs
```diff
@@ -15,5 +20,41 @@ public TagController(ITagService tagService)
...
+        public async Task<ActionResult<bool>> DeleteTag(int? tagId)
+        {
+            if (tagId == null) return Problem("There is no declared tagId");
+
+            var result = await _tagService.DeleteTag(tagId.Value);
+            return Ok(result);
+        }
+
+        [HttpDelete]
+        public async Task<ActionResult<bool>> DeleteTag(int[] tagIds)
```
stutwoby: Why wouldn't we always delete tags by their ID?

Server/JokesOnYou.Web.Api/Controllers/TagController.cs
```diff
@@ -15,5 +20,41 @@ public TagController(ITagService tagService)
...
+        [HttpDelete]
+        public async Task<ActionResult<bool>> DeleteTag(List<TagCreateDto> tags)
+        {
+            if (tags == null) return Problem("There is no declared tags");
+
+            var result = await _tagService.DeleteMultipleTag(tags);
+            return Ok(result);
+        }
+
+        [HttpDelete]
```
HellHunterMax: you could add ("{tagId}") after delete.

Server/JokesOnYou.Web.Api/Controllers/TagController.cs
HellHunterMax: No, if you want to use a DTO (you don't need to because you only need the ID) you have to create a new one for it.

chrisK00: We dont have to return a boolean, its either gonna be an error or nocontent

Server/JokesOnYou.Web.Api/Repositories/TagRepository.cs
```diff
@@ -1,12 +1,73 @@
...
+            _mapper = mapper;
+        }
+        /// <summary>
+        /// Deletes given <paramref name="tag"/>
+        /// </summary>
+        /// <param name="tag"></param>
+        /// <returns></returns>
+        public async Task<bool> Delete(TagCreateDto tag)
+        {
+            var entity = _mapper.Map<Tag>(tag);
```
HellHunterMax: try and move most of the mapping to the service where possible.

Server/JokesOnYou.Web.Api/Controllers/TagController.cs
chrisK00: > Is this the appropriate DTO to delete a tag?

To add on to this. All we need is the id of the item, which will come from the url

Server/JokesOnYou.Web.Api/Repositories/TagRepository.cs
```diff
@@ -1,12 +1,73 @@
...
+
+
+        /// <summary>
+        /// Deletes all given <paramref name="tags"/>
+        /// </summary>
+        /// <param name="tags"></param>
+        /// <returns></returns>
+        public async Task<bool> DeleteRange(List<TagCreateDto> tags)
+        {
+            var entity = _mapper.Map<List<Tag>>(tags);
```
HellHunterMax: I'm not sure when you map a DTO with an id to a object that it is the same object as in the Database, and I don't think it will get deleted that way.

Server/JokesOnYou.Web.Api/Controllers/TagController.cs
```diff
@@ -15,5 +20,41 @@ public TagController(ITagService tagService)
...
+        public async Task<ActionResult<bool>> DeleteTag(int? tagId)
```
chrisK00: dont need this

Server/JokesOnYou.Web.Api/Controllers/TagController.cs
chrisK00: Since this is for multiple tags, consider pluralizing the name

HellHunterMax (changes requested):
If you want to use a Dto for removing. create a new one.
In the end you should only be using the Id so I don't think you should.
2 way's of doing the same thing is not necessary right now.

Right now every member can remove a tag.
you can change it so only the owner of the tag or a Admin can remove it.
But if a tag has likes i don't think the owner should be able to remove it.

Imagine this, some Tag is used by 200 jokes and has 600 likes. The tag gets deleted by a member and we lose all that info.

Server/JokesOnYou.Web.Api/Repositories/Interfaces/ITagRepository.cs
```diff
@@ -1,11 +1,14 @@
...
 using System.Collections.Generic;
-using System.Linq;
 using System.Threading.Tasks;
 
 namespace JokesOnYou.Web.Api.Repositories.Interfaces
 {
     public interface ITagRepository
     {
+        Task<bool> Delete(TagCreateDto tag);
+        Task<bool> Delete(int tagId);
```
chrisK00: This should just recieve a tag (then you can check if its null and throw an exception before removing) and shouldnt return a boolean

Server/JokesOnYou.Web.Api/Repositories/Interfaces/ITagRepository.cs
```diff
@@ -1,11 +1,14 @@
...
+        Task<bool> DeleteRange(List<TagCreateDto> tags);
```
chrisK00: Dont think we need duplicated methods

Server/JokesOnYou.Web.Api/Repositories/TagRepository.cs
```diff
@@ -1,12 +1,73 @@
...
+using JokesOnYou.Web.Api.Repositories.Interfaces;
 using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 
 namespace JokesOnYou.Web.Api.Repositories
 {
-    public class TagRepository : ITagRepository
+    public class TagRepository : UnitOfWork,ITagRepository
```
chrisK00: service should take care of saving not repository

Server/JokesOnYou.Web.Api/Repositories/TagRepository.cs
chrisK00: repo should not contain any business logic. if you wanna read up a bit [code with shadman link](https://codewithshadman.com/repository-pattern-csharp/#:~:text=A%20Repository%20in%20C%23%20mediates,layers%20(like%20Entity%20Framework).&text=Repository%20pattern%20C%23%20is%20a,view%20of%20the%20persistence%20layer.)

chrisK00 (changes requested):

Server/JokesOnYou.Web.Api/Controllers/TagController.cs
oguzhanaksoy: Don't need what ?
Nullable integer or HttpDelete ?

stutwoby (approved):

Server/JokesOnYou.Web.Api/Controllers/TagController.cs
```diff
@@ -1,19 +1,65 @@
...
+            _logger = logger;
+            _userManager = userManager;
+        }
+
+        [HttpDelete]
+        public async Task<IActionResult> DeleteTag(int tagId)
+        {
+            var tag = await _tagService.GetTag(tagId);
+            string userId = _userManager.GetUserId(User);
+            if (tag.UserReplyDTO.Id == userId || User.IsInRole("Admin"))
```
bschaatsbergen: This is a short circuit opperand, if first condition is true we are not evaluating if the used role is "Admin" is this intended?

Server/JokesOnYou.Web.Api/Controllers/TagController.cs
```diff
@@ -1,19 +1,65 @@
...
+            {
+                await _tagService.DeleteTag(tagId);
+                return Ok($"{tagId}");
+            }
+            else
+            {
+                return Unauthorized();
```
bschaatsbergen: Why would it be unauthorized if the first expression in the if condition is not met, I think we should throw a specific error response.

Server/JokesOnYou.Web.Api/Controllers/TagController.cs
bschaatsbergen: This same comment applies to the other implementation

Server/JokesOnYou.Web.Api/Models/BaseOwnedEntity.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace JokesOnYou.Web.Api.Models
+{
+    public abstract class BaseOwnedEntity
+    {
+        public int Id { get; set; }
+        public DateTime Created { get; set; }
```
bschaatsbergen: Created rather sounds like a bool type, perhaps we can rename it to CreationTime or anything else.

HellHunterMax:
As @oguzhanaksoy could not finish I took over.
Changed a lot of stuff. did a big merge from Dev into this.
removed the DeleteRange as I don't see it being used for the app.

Server/JokesOnYou.Web.Api/Models/BaseOwnedEntity.cs
HellHunterMax: removed this model.

Server/JokesOnYou.Web.Api/Controllers/TagController.cs
HellHunterMax: Yes this is intended.
Because if either the User is the owner or the User is an admin s/he can delete the tag (for now).

Server/JokesOnYou.Web.Api/Controllers/TagController.cs
HellHunterMax: But if the User is not the owner nor an Admin s/he's not authorized to do the action ? is this not the intended result ?
We can add a msg to this as well right.

chrisK00 (approved):

HellHunterMax (approved):

PR #118, opened 2021-04-30
Adding Docker Compose

stutwoby:
## What did I do?
We currently have a repo that requires a connection to a database. It's an annoying experience having to set up a local database and mess around with configuring everything. I've added docker support to build both the app and database and spin up containers with connections between them.

## How did I do it?
docker compose, you can add individual docker files to projects, reference them and build them. You can also pull down images from docker hub. In this case, we wanted to reference Postgres which is what I've done. We set up a bridge between the two, pass a connection string and bam, got a lovely database spun up locally without any setup (apart from docker).

Thankfully, we've already got some seeded data, so I've plumbed it in, so we check if the database is created on startup. If not we migrate and populate.


Server/JokesOnYou.Web.Api/Dockerfile
```diff
@@ -0,0 +1,22 @@
+#See https://aka.ms/containerfastmode to understand how Visual Studio uses this Dockerfile to build your images for faster debugging.
```
bschaatsbergen, line 1: We also have a dockerfile in root, is this intended?

docker-compose.override.yml
```diff
@@ -0,0 +1,13 @@
+version: '3.4'
```
bschaatsbergen: Can we solve this with 1 compose?

docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+    depends_on:
+        - "db"
+    ports:
+        - 8000:80
+    environment:
+        DB_CONNECTION_STRING: "host=host.docker.internal;port=5342;database=jokedb;username=jokeuser;password=password"
+    networks:
+        - jokes-on-you-dev
+
+  db:
```
bschaatsbergen, line 26: Maybe add local persistent storage

```
volumes:
      - ./jokesdb_pgdata:/var/lib/postgresql/data
```

docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+    image: postgres:latest
+    restart: always
+    ports:
+        - "5432"
+    environment:
+      POSTGRES_USER: "jokeuser"
```
bschaatsbergen: can remove quotation marks on both user & db env vars

.dockerignore
```diff
@@ -0,0 +1,25 @@
+**/.classpath
+**/.dockerignore
```
bschaatsbergen, line 2: we are ignoring **/.dockerignore but yet committing it

bschaatsbergen:
Nice innovative idea!

docker-compose.override.yml
stutwoby: Can you be clearer? Are you talking about the override file?

docker-compose.yml
stutwoby: Sure, will be in the next commit.

.dockerignore
stutwoby: Good shout, will un-commit

stutwoby: Although we should consider keeping a copy in the repo so when new people download it they don't have to create a new one?

Server/JokesOnYou.Web.Api/Dockerfile
stutwoby: Ah, this is a duplicate. Will remove.

docker-compose.yml
stutwoby: Good shout

docker-compose.override.yml
bschaatsbergen: For the result we want to achieve we don't necessarily need to utilize the override file that docker supports, we can initialize the containers just with 1 compose

.dockerignore
bschaatsbergen: I would keep the .dockerignore in root as it's an important file, otherwise we might fall back into the 'it works on my machine' due to people having different dockerignores overtime

docker-compose.override.yml
stutwoby: Cool, I cleaned it up in the last commit.

.dockerignore
stutwoby: I think if it's a repo wide ignore it should live in the git ignore, docker ignore makes more sense for when we want to ignore docker related files for composing.

bschaatsbergen (approved):

chrisK00 (approved):
Amazing! Looking forward to run this dockerized.

HellHunterMax (approved):
Great explanation on Discord as well thanks for this!

PR #121, opened 2021-05-10
Feature 85 create tag

ofu997:


Server/JokesOnYou.Web.Api/DTOs/TagCreateDto.cs
```diff
@@ -7,5 +7,9 @@ namespace JokesOnYou.Web.Api.DTOs
 {
     public class TagCreateDto
     {
+        // public int Id { get; set; }
+        public string Name { get; set; }
+        public DateTime Created { get; set; } = DateTime.Now;
```
HellHunterMax: remove Created here it will be generated when you map it to the Tag model.

Server/JokesOnYou.Web.Api/DTOs/TagCreateDto.cs
```diff
@@ -7,5 +7,9 @@ namespace JokesOnYou.Web.Api.DTOs
...
+        public int Likes { get; set; }
```
HellHunterMax: remove likes we don't want people to set the amount of likes to start it off with.

Server/JokesOnYou.Web.Api/DTOs/TagReplyDto.cs
```diff
@@ -7,5 +7,9 @@ namespace JokesOnYou.Web.Api.DTOs
 {
     public class TagReplyDto
```
HellHunterMax: add the OwnerId to the reply.

Server/JokesOnYou.Web.Api/Migrations/20210509080507_CreateTag.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.CreateTable(
+                name: "Tags",
+                columns: table => new
+                {
+                    Id = table.Column<int>(type: "INTEGER", nullable: false)
+                        .Annotation("Sqlite:Autoincrement", true),
+                    Name = table.Column<string>(type: "TEXT", nullable: true),
```
HellHunterMax, line 16: should not be nullable

Server/JokesOnYou.Web.Api/Migrations/20210509080507_CreateTag.cs
```diff
@@ -0,0 +1,33 @@
...
+                    Created = table.Column<DateTime>(type: "TEXT", nullable: false),
+                    OwnerId = table.Column<string>(type: "TEXT", nullable: true),
+                    Likes = table.Column<int>(type: "INTEGER", nullable: false)
+                },
+                constraints: table =>
```
HellHunterMax, line 21: this needs another constraint but this code might fix it but might not.
I did this for jokes as well but did it wrong so i have to fix it there to. this is fine for now. don't do anything here.
```
table.ForeignKey(
                        name: "FK_Tags_AspNetUsers_OwnerId",
                        column: x => x.OwnerId,
                        principalTable: "AspNetUsers",
                        principalColumn: "Id",
                        onDelete: ReferentialAction.Restrict);

File: Server/JokesOnYou.Web.Api/Profiles/JokesProfile.cs [hunk 6]
```diff
@@ -15,6 +15,8 @@ public JokesProfile()
             CreateMap<Joke, JokeReplyDto>();
             CreateMap<User, JokeAuthorDto>();
             CreateMap<JokeCreateDto, Joke>();
+            CreateMap<Tag, TagReplyDto>();
+            CreateMap<TagCreateDto, Tag>();
```
HellHunterMax: should create new profile class for these maps.

Server/JokesOnYou.Web.Api/Repositories/Interfaces/ITagRepository.cs
```diff
@@ -1,11 +1,15 @@
-﻿using System;
+﻿using System;//unnecessary?
 using System.Collections.Generic;
-using System.Linq;
+using System.Linq; //unnecessary?
```
HellHunterMax: You can remove all unnecessary using statements.

Server/JokesOnYou.Web.Api/Repositories/Interfaces/ITagRepository.cs
```diff
@@ -1,11 +1,15 @@
...
 using System.Threading.Tasks;
+using JokesOnYou.Web.Api.DTOs;
+using JokesOnYou.Web.Api.Models;
+
```
HellHunterMax, line 5: remove empty line here.

Server/JokesOnYou.Web.Api/Repositories/TagRepository.cs
```diff
@@ -1,12 +1,37 @@
...
+
+        public TagRepository(DataContext context, IMapper mapper)
+        {
+            _context = context;
+            _mapper = mapper;
+        }
+
+        public async Task CreateTagAsync(Tag tag)
+        {
+            try
```
HellHunterMax: remove the try catch here. and make this a oneliner.
If it fails the global exception handler will catch it.

Right now it will just continue as if it did not fail.

Server/JokesOnYou.Web.Api/Services/TagService.cs
```diff
@@ -4,16 +4,40 @@
...
+
+        public async Task<TagReplyDto> CreateTagAsync(TagCreateDto tagCreateDto)
+        {
+            var tag = _mapper.Map<Tag>(tagCreateDto);
+            await _tagRepo.CreateTagAsync(tag);
+
+            var saved = await _unitOfWork.SaveAsync();
+            if (!saved)
+            {
+                throw new AppException("error saving tag");
```
HellHunterMax: You could add some information about the tag in here if you want.

HellHunterMax (changes requested):
Very good job, some minor things to make it perfect. 
Especially the Service is very nicely made!

- [ ] Tag needs an owner (OwnerId) the person creating the tag. should be added in controller using the claim. you can add it to the Dto so you can pass it on to the service.
- [ ] You did the migration (it could use some edits) but you did not `update-database` so the migration is not pushed to the database?

Server/JokesOnYou.Web.Api/Profiles/TagProfile.cs
```diff
@@ -13,6 +13,7 @@ public class TagProfile : Profile
         public TagProfile()
         {
             CreateMap<Tag, TagReplyDto>();
+            CreateMap<TagCreateDto, Tag>();
```
heijmn, line 12: A tag has a OwnerId, how do we make sure that it's correctly set since the Dto doesn't contain it? (Which I think is correct. See: https://github.com/Code2Gether-Discord/JokesOnYou/issues/142 for reference.) 

I guess the person creating the tag is also the owner?

I just saw that HellHunter also mentioned the fact that owner is missing, however I don't like his solution :).

ofu997: Maybe we can set the OwnerId using an Identity field: private readonly UserManager<IdentityUser> _userManager;

HellHunterMax: The Controller already has a way of getting the userId through the Token.
when inside the controller type: `string UserId = ClaimsPrincipalExtension.GetUserId(User);`
That will retrieve the UserId using the token.

HellHunterMax: so a few options.
1. Create a new Dto for the service.
2. add parameters to the service.

anything else ?

heijmn: 3. Make the controller responsible for going from Dto to actual model and pass the model on to the service. 
4. Some CQRS implementation using MediatR (never used it, only saw some videos on it), so I'm not even sure if it will solve the issue.
5. Add the UserManager as a dependency to the TagService, then let the TagService itself get the UserId instead of the controller.

My preference currently: option 2 or 5. But I'd go for the solution provided by a more experienced dev :)

HellHunterMax (approved):
This is a great start to the create tag.
Needs more but we'll add that in a second issue.

Server/JokesOnYou.Web.Api/DTOs/TagCreateDto.cs
```diff
@@ -2,10 +2,13 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using System.ComponentModel.DataAnnotations;
 
 namespace JokesOnYou.Web.Api.DTOs
 {
     public class TagCreateDto
     {
+        [Required()]
```
HellHunterMax: You could remove the () brackets

HellHunterMax (approved):

Niisc (approved):
hi

PR #125, opened 2021-05-14
Feature/front end

HellHunterMax:
# what I made
Made a start on the front-end.
- A list of jokes will be retrieved from the Api (get all jokes).
- You can click on one of the jokes to view a single joke.
- Added cors to the backend.
With cors the website is able to make the Api calls.
The cors I implemented is not safe at all so needs to be changed later.

### Side notes
- The link for the api call is constant at the moment and may be needed to change depending on your setup.
- I cant get docker to work for me so my port was 5000 for the Api calls.
- No css is used yet no styling at all.
- if you want to add to this feel free to do so!

### How to start it:
1. right click on client.
2. press `open in terminal`
3. type in `npm start`
4. it should tell you in the terminal where to open the browser on. (for me it was: `http://localhost:4200/` )
5. start up the backend by pressing `JokesOnYou.Web.Api` in the top of visual studio.

### how it works
app.module.ts is like the main hub for angular, this is where you tell angular what components are used for the website.
you import all the components there and add them to the declarations.
Paths you use are given using:
```
RouterModule.forRoot([
      { path: '', component: JokesComponent },
      { path: 'joke/:id', component: JokeComponent }
    ])
```
in this case App.Component.ts is the main page everything you want to have should be on there.
as you can see:
```
<app-navbar></app-navbar>
<div class="container">
  <router-outlet></router-outlet>
</div>
```
The `navbar` is on top and then comes the chosen pages that come from the router see the routerModule.

Next to the navbar i created 2 more components:
`Jokes.component` = a list of all jokes.
`joke.component` = a view for a single joke.

#### jokes.component
I created a interface file for joke (model) that is being imported at the top as Joke
Declare a list of jokes: `jokes?: Joke[];`
declared the url to make the call to (this should come from somewhere else `private jokesUrl = 'http://localhost:5000/api/jokes';`
inside the constructor set http client for making calls: `constructor(private http: HttpClient) { }`
made a method that gets a list of all jokes and assigns it to jokes 'getJokes(): void {this.http.get<Joke[]>(this.jokesUrl).subscribe(jokes => this.jokes = jokes);}'
and finally on initialization it will make the call: `ngOnInit() { this.getJokes() }`

### Documentation I used
[Angular getting started](https://angular.io/start)
[enabling cors](https://stackoverflow.com/questions/44379560/how-to-enable-cors-in-asp-net-core-webapi)
[Tour of Heroes](https://angular.io/tutorial)


ofu997:
Good starting point for endpoints '/' and '/joke/${id}'. We need to take note we also need to run npm install and install angular cli before running npm start

chrisK00 (approved):

PR #138, opened 2021-05-20
Added Author name to GetJoke endpoint #127

### LINKED ISSUE #127: Endpoint: Get JokeAuthor
### Usecase:
when viewing single joke you would like to see the name of the author.

Maybe we should change the getJoke endpoint to include the author name.
getAlljokes does not need author name right now.

## how
create a get endpoint for JokeAuthor
return a JokeAuthorDto for the author in question.
`api/jokes/author/{id}`

heijmn:
Updated controller, service and repository to use the new JokeWithAuthorReplyDto

https://github.com/Code2Gether-Discord/JokesOnYou/issues/127


heijmn:
@HellHunterMax, I gave this a try but I'm not sure whether this is the correct way forward. Can you take a look at this?

I chose to add the Author.Name to the GetJoke endpoint instead of creating a specific GetAuthor endpoint. This way the client needs to do only one round-trip. However I'm not sure if I like the inheritance of the JokeReplyDto. Maybe it would be better to create a separate Dto for a JokeWithAuthor where the Author is a nested object. I'm not sure though :)

HellHunterMax:
First off, Great job!

second, @chrisK00 can you explain why we changed the models from, for example, User into UserId ?
Because I am seeing more and more things that I don't like about it.

Third, the Inheriting of JokeReplyDto is a nice one to keep it `dry` I have to say. I have absolutely no clue if its good or not :P

fourth, we will know which API calls we need when we have more frontend.
But think about it this way. In what situation would you want to get a joke without the name of the author ?
Maybe when you are looking at the jokes you created yourself? so that's one example on how we could use that endpoint.
but still its not bad to have the extra Author name it might be more hassle to handle two endpoints.

chrisK00:
> second, @chrisK00 can you explain why we changed the models from, for example, User into UserId ?
> Because I am seeing more and more things that I don't like about it.

This is a matter about design and ref by id (which is a advanced concept but its always prefered). For example: Why would you get a users password when you want to get a joke? 

So if we have a UserId we basically have a Foreign Key which can be used to either identify that the user owns the joke, or to get a user. But we do not include the user with the joke (unless for some reason this is what you want).

heijmn:
@chrisK00, @HellHunterMax, I've removed the User reference and made the service fetch the user based on the AuthorId and then append it to the Dto. This way we follow the DDD rule of not having a direct reference to another aggregate.

However now we retrieve the whole user just to get the name (because I'm using UserManager and I don't know if it's possible to use ProjectTo or something similar to only retrieve certain fields). Not sure if this is an issue though.

BTW: The build seems to fail ('User' does not contain a definition for 'Name' and no accessible extension method 'Name' accepting a first argument of type 'User' could be found (are you missing a using directive or an assembly reference?)) although it's working on my machine...

Server/JokesOnYou.Web.Api/Controllers/JokesController.cs
```diff
@@ -54,7 +54,7 @@ public async Task<ActionResult<JokeReplyDto>> UpdateJoke(JokeUpdateDto jokeUpdat
 
         [AllowAnonymous]
         [HttpGet("{id}")]
-        public async Task<ActionResult<JokeReplyDto>> GetJoke(int id)
+        public async Task<ActionResult<JokeWithAuthorReplyDto>> GetJoke(int id)
```
chrisK00: I think this should just be JokeDto, we can reuse this model as we always want the author. Naming it both reply and dto is not needed (yes i know we have it on some places) but we have an issue for renaming theese to just responses instead

Server/JokesOnYou.Web.Api/DTOs/JokeWithAuthorReplyDto.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace JokesOnYou.Web.Api.DTOs
+{
+    public class JokeWithAuthorReplyDto : JokeReplyDto
```
chrisK00: a dto Inhereting from another dto is generally not a good idea, since they should look pretty different otherwise we can reuse the same

Server/JokesOnYou.Web.Api/Repositories/Interfaces/IJokesRepository.cs
```diff
@@ -13,6 +13,6 @@ public interface IJokesRepository
         Task<Joke> GetJokeByIdAsync(int id);
         Task CreateJokeAsync(Joke joke);
         Task<bool> DoesJokeExist(string normalizedPremise, string normalizedPunchline);
-        Task<JokeReplyDto> GetJokeDtoAsync(int id);
+        Task<JokeWithAuthorReplyDto> GetJokeDtoAsync(int id);
```
chrisK00: Wrong return type or wrong method name

Server/JokesOnYou.Web.Api/Repositories/JokesRepository.cs
```diff
@@ -30,7 +30,12 @@ public JokesRepository(DataContext context, IMapper mapper)
         public async Task<IEnumerable<JokeReplyDto>> GetAllJokeDtosAsync() => await _context.Jokes.ProjectTo<JokeReplyDto>(_mapper.ConfigurationProvider).ToListAsync();
         public async Task<Joke> GetJokeByIdAsync(int id) => await _context.Jokes.FirstOrDefaultAsync(x => x.Id == id);
 
-        public async Task<JokeReplyDto> GetJokeDtoAsync(int id) => await _context.Jokes.ProjectTo<JokeReplyDto>(_mapper.ConfigurationProvider).FirstOrDefaultAsync(j => j.Id == id);
+        public async Task<JokeWithAuthorReplyDto> GetJokeDtoAsync(int id) => 
+            await _context
+            .Jokes
```
chrisK00: _context.Jokes can be on the same line

Server/JokesOnYou.Web.Api/Services/JokesService.cs
```diff
@@ -47,7 +46,7 @@
             var saved = await _unitOfWork.SaveAsync();
             if (!saved)
             {
-                if(await _jokesRepo.GetJokeDtoAsync(joke.Id) == null)
```
chrisK00, line 50: You can remove this since: If the save failed, theres no reason to try and get the item from the db thats just another request made to the db

chrisK00 (changes requested):
Very well done. Just some small improvements

Server/JokesOnYou.Web.Api/Controllers/JokesController.cs
heijmn: The idea of the issue was to only include AuthorName when you retrieve a specific Joke. For example when retrieving a list of Jokes, maybe you don't want to include the Authors (could potentially be a lot of different authors).

With that in mind, do you still think it should be the same Dto? If yes, will we just leave AuthorName empty when retrieving a list of Jokes or also populate the AuthorName in that case?

Server/JokesOnYou.Web.Api/Services/JokesService.cs
heijmn: This wasn't my code. My OCD just had to add that space after the if statement. But following the boy-scout-rule I will fix it anyhow :)

Server/JokesOnYou.Web.Api/Controllers/JokesController.cs
chrisK00: always populate i suppose, since it does make sense to always have the author username presented to the user

Server/JokesOnYou.Web.Api/Services/JokesService.cs
```diff
@@ -66,10 +63,15 @@ private static void TrimAndNormalizeJokeCreateDto(JokeCreateDto jokeCreateDto)
             jokeCreateDto.NormalizedPunchline = jokeCreateDto.Punchline.ToUpper();
         }
 
-        public async Task<IEnumerable<JokeReplyDto>> GetAllJokeDtosAsync()
+        public async Task<IEnumerable<JokeDto>> GetAllJokeDtosAsync()
         {
             var jokeDtos = await _jokesRepo.GetAllJokeDtosAsync();
 
+            foreach (var jokeDto in jokeDtos)
```
heijmn, line 71: There must be better ways to handle this...?

Server/JokesOnYou.Web.Api/DTOs/JokeDto.cs
```diff
@@ -1,17 +1,14 @@
...
 
 namespace JokesOnYou.Web.Api.DTOs
 {
-    public class JokeReplyDto
+    public class JokeDto
     {
         public int Id { get; set; }
         public string Premise { get; set; }
         public string Punchline { get; set; }
         public string AuthorId { get; set; }
```
heijmn: I'd rather have Author Id and Name as a nested object (something like JokeAuthorDto). But when I remove the AuthorId from JokeDto, then it will not be available for me to retrieve the name in the second query...

I'm lost :)

Server/JokesOnYou.Web.Api/Services/JokesService.cs
```diff
@@ -47,15 +46,13 @@ public async Task<JokeReplyDto> CreateJokeAsync(JokeCreateDto jokeCreateDto)
...
-                if(await _jokesRepo.GetJokeDtoAsync(joke.Id) == null)
-                {
-                    throw new AppException("Error with saving the Joke.");
-                }
+                throw new AppException("Error with saving the Joke.");
             }
 
-            var jokeReplyDto = _mapper.Map<JokeReplyDto>(joke);
+            var jokeDto = _mapper.Map<JokeDto>(joke);
+            jokeDto.AuthorName = user.Name;
```
heijmn: For me this doesn't "feel right"...

heijmn:
Well I did some more changes (to always return AuthorName) but I don't like the outcome. Maybe someone else could try to tackle the issue as well from a clean perspective and then we can discuss why a certain solution is better or worse :).

HellHunterMax:
> > second, @chrisK00 can you explain why we changed the models from, for example, User into UserId ?
> > Because I am seeing more and more things that I don't like about it.
> 
> This is a matter about design and ref by id (which is a advanced concept but its always prefered). For example: Why would you get a users password when you want to get a joke?

You don't ? you use mapping and ProjectTo to choose which items you retrieve from the database.


> So if we have a UserId we basically have a Foreign Key which can be used to either identify that the user owns the joke, or to get a user. But we do not include the user with the joke (unless for some reason this is what you want).

The User inside the joke model is always just a reference to the user.
and you can always choose how to retrieve it.

But if two calls is just as good and fast as 1 call then lets keep it like this and make two calls.

Server/JokesOnYou.Web.Api/Services/JokesService.cs
HellHunterMax: This is fine for now. there is a second way.
GetUserReplyAsync then you don't get all fields from the user.

Server/JokesOnYou.Web.Api/Services/JokesService.cs
HellHunterMax: I thought we agreed that its possible some other method saved this change?
then this would fail and require a second check?

Server/JokesOnYou.Web.Api/DTOs/JokeDto.cs
HellHunterMax: A joke Always has an Author and will always be used imo.
so this is good.

Server/JokesOnYou.Web.Api/Services/JokesService.cs
```diff
@@ -100,12 +102,34 @@ public async Task<JokeReplyDto> UpdateJoke(JokeUpdateDto jokeUpdateDto)
...
         }
 
-        public Task<JokeReplyDto> GetJokeDtoAsync(int id)
+        private async Task AddAuthorToJoke(JokeDto jokeDto)
         {
-            return _jokesRepo.GetJokeDtoAsync(id);
+            var user = await _userRepository.GetUserAsync(jokeDto.AuthorId);
+            if (user == null)
+            {
+                throw new AppException($"No user with this id:\"{jokeDto.AuthorId}\" found in the Database.");
```
HellHunterMax: I'm not sure we should throw here. if one out of 100 jokes has a bad Id then the whole method goes corrupt.
Make a console.Log and set the AuthorName to something like notfound or something ?

Server/JokesOnYou.Web.Api/Services/JokesService.cs
HellHunterMax: use AddAuthorToJoke(jokeDto);

HellHunterMax (commented):
I still think changing UserID to User would make for a lot less code and better maintainability .

@chrisK00 look at all this code that could have been left out and all these separate calls if we just use User in stead of UserId.
Then i read some more about it and i'm thinking we need the models in here.
https://udidahan.com/2009/01/24/ddd-many-to-many-object-relational-mapping/
and
https://enterprisecraftsmanship.com/posts/link-to-an-aggregate-reference-or-id/

I don't know if the articles are any good but its seems to make sense. and this code right now does not for me. please change my mind ;)

chrisK00:
> I still think changing UserID to User would make for a lot less code and better maintainability .
> 
> @chrisK00 look at all this code that could have been left out and all these separate calls if we just use User in stead of UserId.
> Then i read some more about it and i'm thinking we need the models in here.
> https://udidahan.com/2009/01/24/ddd-many-to-many-object-relational-mapping/
> and
> https://enterprisecraftsmanship.com/posts/link-to-an-aggregate-reference-or-id/
> 
> I don't know if the articles are any good but its seems to make sense. and this code right now does not for me. please change my mind ;)

Well for starters, why would you intentionally make a self referencing loop? User has jokes....joke has a user... That user has jokes... The joke has a user and so on.

Think of how sql does it, and no sql does it. Theres no database that makes a self referencing loop intentionally, even Entity framework is smart enough to not do that. Under the hood it uses Id.

Just saying the relationship shouldnt be bidirectional because it doesnt make sense, not in the real world either. A joke cant own a user. Its owned by a user.

Ddd isnt about having short or fast code, its about maintainability and solving design problems by making them more towards how things are in life. Now im not saying Ddd solves all our issues, but modelling entities using reference by id is a standard because like i said otherwise your just making a self referencing loop. I have heard that some people stay away from it if they have Juniors on their teams which makes sense but thats just if they use EF.

chrisK00:
> > > second, @chrisK00 can you explain why we changed the models from, for example, User into UserId ?
> > > Because I am seeing more and more things that I don't like about it.
> > 
> > 
> > This is a matter about design and ref by id (which is a advanced concept but its always prefered). For example: Why would you get a users password when you want to get a joke?
> 
> You don't ? you use mapping and ProjectTo to choose which items you retrieve from the database.
> 
> > So if we have a UserId we basically have a Foreign Key which can be used to either identify that the user owns the joke, or to get a user. But we do not include the user with the joke (unless for some reason this is what you want).
> 
> The User inside the joke model is always just a reference to the user.
> and you can always choose how to retrieve it.
> 
> But if two calls is just as good and fast as 1 call then lets keep it like this and make two calls.

If you are a Ddd fan, original repository fan and cared about maintainability you wouldnt even be allowed to use projectto inside your repository. Because dtos are not located inside that tier so it wouldnt even be possible for doing that.

Like your saying you have a choice to include it, then whats the point polluting your repository with different methods and polluting your models with unnecesary complicated relationships.

Server/JokesOnYou.Web.Api/Services/JokesService.cs
heijmn: > use AddAuthorToJoke(jokeDto);

I already have the user in the scope of this method. In the method "AddAuthorToJoke(jokeDto)" it's retrieved again by Id. It would be an unnecessary call to the UserManager.

heijmn:
I don't have enough experience working on large projects with multiple teams, neither do I know a lot about DDD. So please don't read below as: "I don't agree", I just want to better understand the stuff 

> Well for starters, why would you intentionally make a self referencing loop? User has jokes....joke has a user... That user has jokes... The joke has a user and so on.

Can you elaborate a bit on this? I don't see what's wrong with a self referencing loop when modeling objects. It's something which exists in the real world as well. 
- I have a wife, my wife has a husband, that's me again having a wife... 
- I have two cars, each car has an owner, that's me having two cars again...

From a DDD maintainability perspective I understand that you don't want to provide access to the related object (A) from an object (B) that should not update A. From a transactional point I understand that you want to make sure aggregates are handled in separate transactions. I guess you have to experience the advantages to understand that they weigh up to the disadvantages.

> Like your saying you have a choice to include it, then whats the point polluting your repository with different methods and polluting your models with unnecesary complicated relationships.

I understand the comment wrt. the repository and keeping it clean. But the relation which you call unnecessary is actually there... It's just not a direct reference to an object but it's abstracted down to an Id... This doesn't change anything about the fact that there actually IS a relation between the two.

chrisK00:
> * I have a wife, my wife has a husband, that's me again having a wife...
Do you own your wife, and your wife owns you? That sounds a bit strange to me... I think both of you rather have something unique that references the other (id) rather than you carrying each other everywhere.
> * I have two cars, each car has an owner, that's me having two cars again...
Since when does a car own a human? (unless Tesla has evolved very quickly i dont think this is possible). A car has a owner, it doesnt own a owner. It has a ownerId, not an actual owner.

> I understand the comment wrt. the repository and keeping it clean. But the relation which you call unnecessary is actually there... It's just not a direct reference to an object but it's abstracted down to an Id... This doesn't change anything about the fact that there actually IS a relation between the two.
Im not removing the relationship nor implying there is none, im simply making it look more realistic. For example if we have a look at a User, would you normaly Get a user with his Jokes - Or Get a  user, then go on to his Jokes tab (with this you wouldnt even have the list of jokes on the user, EF provides so easy ways of making stuff like this and if you go on any website this is how they do it because imagine your user class having a billion lists).

Server/JokesOnYou.Web.Api/DTOs/JokeDto.cs
```diff
@@ -2,12 +2,12 @@
 
 namespace JokesOnYou.Web.Api.DTOs
 {
-    public class JokeReplyDto
+    public class JokeDto
     {
         public int Id { get; set; }
         public string Premise { get; set; }
         public string Punchline { get; set; }
-        public string AuthorId { get; set; }
```
chrisK00, line 10: I think we just need author id and author username so we could just have them as inline properties

Server/JokesOnYou.Web.Api/Profiles/JokesProfile.cs
```diff
@@ -8,8 +8,8 @@ public class JokesProfile : Profile
     {
         public JokesProfile()
         {
-            CreateMap<Joke, JokeReplyDto>();
-            CreateMap<User, JokeAuthorDto>();
+            CreateMap<Joke, JokeDto>().ForMember(dest => dest.Author, x => x.MapFrom(scr => new JokeAuthorDto() { Id = scr.AuthorId }));
```
chrisK00: Think you meant src - source

Server/JokesOnYou.Web.Api/Services/JokesService.cs
```diff
@@ -100,12 +103,34 @@ public async Task<JokeReplyDto> UpdateJoke(JokeUpdateDto jokeUpdateDto)
...
+
+            return jokeDto;
         }
 
-        public Task<JokeReplyDto> GetJokeDtoAsync(int id)
+        private async Task AddAuthorToJoke(JokeDto jokeDto)
         {
-            return _jokesRepo.GetJokeDtoAsync(id);
+            var user = await _userRepository.GetUserReplyAsync(jokeDto.Author.Id);
+            if (user == null)
```
chrisK00: Is the user property on a joke "null" in our db? I think a  joke has to have a user so no need to check for null

chrisK00 (changes requested):
Gj!

Server/JokesOnYou.Web.Api/DTOs/JokeDto.cs
HellHunterMax: We could but its not part of the Joke so I think it should have its own Dto.
And Author Dto could be used for more than just the joke and we might add more then just the UserName.

chrisK00 (approved):
🥇

heijmn:
@HellHunterMax, thanks for picking this up further.

HellHunterMax (approved):

PR #144, opened 2021-05-20
Feature/136 account service

### LINKED ISSUE #136: AuthService for login/register
For #135 

Create an auth service that makes use of [httpclient](https://angular.io/guide/http) in order to make requests to our api. 

For the login we need to persist our users information somehow in case the page is refreshed. The easiest way is using localstorage.

`localStorage.setItem('key', value);`
`localStorage.getItem('key')`

chrisK00:
1. I moved folders in to app folder and updated namespaces
2. Since we have strict mode enabled, i had to make all User properties optional in order to make use of type safety.
3. setCurrentUser() if we have a user in localstorage, we can just set the currentUser
4. login method takes in a type of any, unsure if you want to have a request model for this instead.

closes #136


stutwoby (approved):

Client/src/app/_services/account.service.ts
```diff
@@ -0,0 +1,36 @@
...
+import { User } from '../_models/user';
+import { map } from 'rxjs/operators';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class AccountService {
+  private accountUrl = environment.apiUrl + 'account';
+  private currentUser = new ReplaySubject<User>(1);
+  currentUser$ = this.currentUser.asObservable();
```
HellHunterMax, line 15: Can you explain how this works and what it means ?

Client/src/app/_services/account.service.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  constructor(private http: HttpClient) { }
+
+  login(model: any) {
```
HellHunterMax: Add UserLoginDto here ?

HellHunterMax (approved):
this looks fkking awesome and very clean.
Makes me want to learn more Angular.
- [x] I guess we could add the UserLoginDto as model ?

Client/src/app/_services/account.service.ts
chrisK00: https://github.com/Code2Gether-Discord/JokesOnYou/wiki/Account-Service-in-Angular-Client

PR #160, opened 2021-05-24
Added thing to make sure that email and username are correct when Reg…

### LINKED ISSUE #149: Refactor: Account: change login to use either email or username.
Right now The UserLoginDto has UserName and Email.
Change this into one input. 
then when Looking up the user check the UserName and Email fields if the user exists.

Niisc:
…istering so that people don't put emails as usernames and, added ability to login both with username and email


N1c0z:
hi gamers

chrisK00 (approved):

HellHunterMax (approved):
Looks Great!

PR #163, opened 2021-06-01
Refactor for Dto's and Repo

### LINKED ISSUE #108: Move User profiles inside Jokes to User profiles
left them in Jokes because users profile was not created yet.

### LINKED ISSUE #116: Refactor: Change usages of returned Tasks to be awaited
Reference discussion [here](https://github.com/Code2Gether-Discord/JokesOnYou/pull/113#discussion_r620478971).

Essentially, update all usages where `Task`'s are returned without being awaited, to await them.

Example
```c#
public Task DoThing() => return Task.Delay(1);
// should be
public async Task DoThing() => await Task.Delay(1);
```

Additionally, if there is a way to enforce this in a lint step, that could be investigated.

### LINKED ISSUE #120: Convert the DTO concept into Request/Response models
Currently the sln is using DTO Models which are quite vaguely implemented, I would suggest refactoring these to Request/Response models.

A controller serializes to a XRequestModel and returns a XResponseModel.

### LINKED ISSUE #141: Remove NormalizedPremise and NormalizedPunchline from JokeCreateDto
Currently the JokeCreateDto contains the fields NormalizedPremise and NormalizedPunchline although they shouldn't be set by the client API. Even if the client set's them, the server will ignore the values. Therefore they shouldn't be part of the Dto in the first place.

This requires a rewrite of the service layer.

### LINKED ISSUE #150: Refactor: Account: Login should use mapper
self explanatory.
Login should use mapper and its not right now.

HellHunterMax:
closes #120


HellHunterMax:
Just changed this to also close #116 
as its one big refactor.

HellHunterMax:
closes #150 
Added mapper to UserLogin
Register still does it manually since it seems better.

HellHunterMax:
closes #108 as well i moved the profiles into right files.

HellHunterMax:
Added a Duplicate check for UpdateJoke

HellHunterMax:
closes #141

HellHunterMax:
Merged Dev into this and then tested all endpoints again, they still work.

chrisK00 (approved):

PR #165, opened 2021-06-15
Feature/91 filter feature

### LINKED ISSUE #91: feature: Get jokes by Filter
add code for Get jokes by Filter

- [ ] http get (filter model?)

### LINKED ISSUE #126: Endpoint: Get User With Jokes
### usecase:
Guest:
1. view single joke.
2. press on joke Author
3. see Author (user) page including a list of jokes form that author.

unbekanntunity:
- Add filter feature for author, amounts of likes and dislikes and the uploaddate


Server/JokesOnYou.Web.Api/Extensions/LINQExtensions.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace JokesOnYou.Web.Api.Extensions
+{
+    public static class LINQExtensions
+    {
+        public static IEnumerable<TSource> WhereIf<TSource>(this IEnumerable<TSource> source, Func<TSource, bool> predicate, bool value)
```
PilgrimViis, line 10: You don't use this method anywhere. I suggest deleting the file. Plus I find the semantics of this method a bit misleading.

Server/JokesOnYou.Web.Api/Filters/JokesFilter.cs
```diff
@@ -0,0 +1,25 @@
...
+        public int Likes { get; set; }
+        public int MinLikes { get => minLikes; set => minLikes = (value > MaxLikes) ? MaxLikes : value; }
+        public int MaxLikes { get => maxLikes; set => maxLikes = (value < MinLikes) ? MinLikes : value; }
+        public int Dislikes { get; set; } = 0;
+        public int MinDislikes { get => minDislikes; set => minDislikes = (value > MaxDislikes) ? MaxDislikes : value; }
+        public int MaxDislikes { get => maxDislikes; set => maxDislikes = (value < MinDislikes) ? MinDislikes : value; }
+        public DateTime UploadDate { get; set; }
+        public DateTime MinUploadDate { get => minUploadDate; set => minUploadDate = (value > MaxUploadDate) ? MaxUploadDate : value; }
+        public DateTime MaxUploadDate { get => maxUploadDate; set => maxUploadDate = (value < MinUploadDate) ? MinUploadDate : value; }
+    }
```
PilgrimViis, line 24: Do we really need this kind of sorting? I mean as a user, I want to sort by likes, but why do I want to sort posts with more than 10 dislikes and less than 100 likes? You also can use uint for dislikes and likes.  We can filter wrong values using asp net attributes or fluentvalidation library.

Server/JokesOnYou.Web.Api/Repositories/Interfaces/IJokesRepository.cs
```diff
@@ -14,5 +15,6 @@ public interface IJokesRepository
         Task CreateJokeAsync(Joke joke);
         Task<bool> DoesJokeExist(string normalizedPremise, string normalizedPunchline);
         Task<JokeDto> GetJokeDtoAsync(int id);
+        Task<IEnumerable<JokeDto>> GetFilteredJokeDtos(JokesFilter jokeFilter);
```
PilgrimViis, line 18: This interface is inconsistent with naming. Sometimes method's names ends with Async and sometimes not.

Server/JokesOnYou.Web.Api/Filters/JokesFilter.cs
PilgrimViis: Also, I'm suggesting using something like BaseFilter with properties **Limit** and **Offset** for pagination and **SortBy**  and **SortDesc**  for sorting. All other filters could implement BaseFilter and add custom filtering properties like NumberOfLikes and Authors.

HellHunterMax: Example from @PilgrimViis :
```
/// <summary>
    /// Special filter for book entities
    /// </summary>
    public class BookFilter : Filter
    {
        /// <summary>
        /// We need to implement special filter. We want to give a user a way to search for specific year 
        /// when books were published.
        /// </summary>
        public int? PublishedAt { get; set; }
    }
    /// <summary>
    /// General filter
    /// </summary>
    public class Filter
    {
        /// <summary>
        /// General search through all meaningfull parameters.
        /// </summary>
        public string Search { get; set; }

        /// <summary>
        /// How many entities should be in Response. If we have 100 books, we but 10 in Count, the API will return 10.
        /// Should be applied after search and sorting
        /// </summary>
        public int Count { get; set; } = 10;

        /// <summary>
        /// How many entities should be skipped. For example, we've already showed 3 pages (or 30 entities with infinite scroll)
        /// and how we need to load entities from 31 to 41. So we make Offset equal to 30. Should be applied after search and sorting
        /// </summary>
        public int Offset { get; set; } = 0;

        /// <summary>
        /// Sort by which field. We can pass "author" and the results will be sorted by autor's names.
        /// </summary>
        public string SortBy { get; set; }

        /// <summary>
        /// If we need Descending sorting we might pass "true" here
        /// </summary>
        public bool SortDesc { get; set; } = false;
    }
    ```

chrisK00: Good tip! Filter and Paging should be two separate classes tho

PR #169, opened 2021-06-27
feature/84-User-Saved-Joke

### LINKED ISSUE #84: Feature: User Saved Joke
Change the models To include Save Feature.

- [ ] Joke Model
- [ ] User Model
- [ ] Dto's
- [ ] http Post for saving (linked to the User)
- [ ] http Delete to delete the save

Niisc:


Server/JokesOnYou.Web.Api/Controllers/SavedJokeController.cs
```diff
@@ -11,5 +20,17 @@ public SavedJokeController(ISavedJokeService savedJokeService)
...
+        [HttpPost("{id}")]
+        public async Task<ActionResult> AddSavedJoke(int id)//id here is jokeid
+        {
+            await _savedJokeService.AddSavedJoke(id, ClaimsPrincipalExtension.GetUserId(User));
+            return NoContent();
+        }
+        [HttpGet]
+        public async Task<ActionResult<IEnumerable<SavedJokeReplyDto>>> GetSavedJokesByUserId()
+        {
+            var stuff = await _savedJokeService.GetSavedJokesByUserId(ClaimsPrincipalExtension.GetUserId(User));
```
chrisK00: stuff? xD

Server/JokesOnYou.Web.Api/Models/SavedJoke.cs
```diff
@@ -1,12 +1,15 @@
...
+using System.ComponentModel.DataAnnotations;
 
 namespace JokesOnYou.Web.Api.Models
 {
     public class SavedJoke
     {
         public int Id { get; set; }
+        [Required]
         public int JokeId { get; set; }
-        public DateTime SavedDate { get; set; } = DateTime.Now;
```
chrisK00, line 9: whats the reason for removing the = DateTime.Now? (should be DateTime.UtcNow fyi)

Server/JokesOnYou.Web.Api/Repositories/SavedJokeRepository.cs
```diff
@@ -1,8 +1,30 @@
...
+        private readonly DataContext _context;
+        private readonly IMapper _mapper;
+
+        public SavedJokeRepository(DataContext context, IMapper mapper)
+        {
+            _context = context;
+            _mapper = mapper;
+        }
+
+        public async Task AddSavedJoke(SavedJoke savedJoke) => await _context.SavedJokes.AddAsync(savedJoke).AsTask();
```
chrisK00: dont need to append .AsTask since your using await

Server/JokesOnYou.Web.Api/Repositories/SavedJokeRepository.cs
```diff
@@ -1,8 +1,30 @@
...
+        public async Task<IEnumerable<SavedJokeReplyDto>> GetSavedJokesByUserId(string id) => await _context.SavedJokes.Where(x => x.UserId == id).ProjectTo<SavedJokeReplyDto>(_mapper.ConfigurationProvider).ToListAsync();
```
chrisK00: split this into atleast 2 lines

chrisK00 (changes requested):

Server/JokesOnYou.Web.Api/Services/SavedJokeService.cs
```diff
@@ -1,15 +1,42 @@
...
+            _unitOfWork = unitOfWork;
+        }
+
+        public async Task AddSavedJoke(int jokeId, string userId)
+        {
+            await _savedJokeRepo.AddSavedJoke(new SavedJoke()
+            {
+                JokeId = jokeId,
+                UserId = userId,
+                SavedDate = System.DateTime.Now
```
HellHunterMax: SavedDate is not needed as it automatically done when creating a new SavedJoke.

HellHunterMax (approved):
Good job Really well written!
I'm not sold about the own controller/Service/Repo.
Repo I understand because it has its own database entry.
Service and Controller could be put under User imo.

HellHunterMax:
You have some conflicts to resolve too. Please Pull dev and resolve them.

Niisc:
> 
> 
> Good job Really well written!
> I'm not sold about the own controller/Service/Repo.
> Repo I understand because it has its own database entry.
> Service and Controller could be put under User imo.

Hmm, imo its more organized like this
@chrisK00 opinions?

chrisK00:
> > Good job Really well written!
> > I'm not sold about the own controller/Service/Repo.
> > Repo I understand because it has its own database entry.
> > Service and Controller could be put under User imo.
> 
> Hmm, imo its more organized like this
> @chrisK00 opinions?

Splitting up the responsibilites is important to follow SRP and provides more decoupling, easier classes to navigate. If we dont do something like this we need to switch over to MediatR. This approach also doesnt break OCP

We shouldnt stick all our logic in one class, and the responsibilites are a bit different. If we want to have more features regarding our jokes, we wouldnt put them in the jokes controller because we would run in to a OverFlowException (bad joke)

So if we think about our consumers
they get jokes on api/jokes
but to get saved jokes they get them on api/savedjokes - i think this is totally fine but the alternative is to simply extend the jokes controller. So the route would become api/jokes/saved which is a bit easier to keep track on but at the same time we still split up the classes.

The important part is that we dont nest to much because that will cause a very ugly url, but in this case i cannot see what else we would nest after the saved.

HellHunterMax:
> > > Good job Really well written!
> > > I'm not sold about the own controller/Service/Repo.
> > > Repo I understand because it has its own database entry.
> > > Service and Controller could be put under User imo.
> > 
> > 
> > Hmm, imo its more organized like this
> > @chrisK00 opinions?
> 
> Splitting up the responsibilites is important to follow SRP and provides more decoupling, easier classes to navigate. If we dont do something like this we need to switch over to MediatR. This approach also doesnt break OCP
> 
> We shouldnt stick all our logic in one class, and the responsibilites are a bit different. If we want to have more features regarding our jokes, we wouldnt put them in the jokes controller because we would run in to a OverFlowException (bad joke)
> 
> So if we think about our consumers
> they get jokes on api/jokes
> but to get saved jokes they get them on api/savedjokes - i think this is totally fine but the alternative is to simply extend the jokes controller. So the route would become api/jokes/saved which is a bit easier to keep track on but at the same time we still split up the classes.
> 
> The important part is that we dont nest to much because that will cause a very ugly url, but in this case i cannot see what else we would nest after the saved.

Thank you for the great explanation! This is where I need to learn.

Server/JokesOnYou.Web.Api/Migrations/20210626132628_SavedJokes.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+using Microsoft.EntityFrameworkCore.Migrations;
+using Npgsql.EntityFrameworkCore.PostgreSQL.Metadata;
```
HellHunterMax, line 3: Not using Postgress anymore

HellHunterMax (changes requested):
I think you have to revert the Migration because its still set to Postgres Database.
I get a build error because of that as well.
Maybe easiest way would be to start over from dev branching out, then copy your new code from here over to the new branch ?

Would be cool if we could remove a saved joke ?

Looks good though.

Server/JokesOnYou.Web.Api/Migrations/20210626132628_SavedJokes.cs
Niisc: oh lol, something got screwed in the dev with saved joke merge

Niisc: thank you for pointing it out as i missed it

PR #170, opened 2021-08-20
Added a new component for login page

PilgrimViis:


PilgrimViis:
I took bye's registration page and have updated it. I've tried to merge his branch into mine, but I've failed =(  I've made a redesign if you don't mind. I want to make "Login" and "Register" buttons disabled, until the whole form would be valid. It doesn't look good with just border shadowing. Also, I've tried to take byes slider, but it looked awful on mobile, so I've switched to bootstrap switch.
image

Client/src/app/_models/loginInfo.ts
```diff
@@ -0,0 +1,4 @@
+export interface LoginInfo {
+  email?: string;
```
chrisK00: I think you are currently able to login either using email or username so having this named username might be more appropiate

Client/src/app/_models/registerInfo.ts
```diff
@@ -0,0 +1,8 @@
+export interface RegisterInfo {
+  username?: string;
+  dateOfBirth?: string;
```
chrisK00: Date type

chrisK00 (changes requested):
Very readable code :tada:

chrisK00 (approved):

PR #172, opened 2021-09-11
created a variable called useSQLite which can be changed depending on…

### LINKED ISSUE #164: Make docker optional
Docker makes more sense in a dev environment, but right now it is just slowing down the development process alot.

Make docker optional! One approach would be to have another connection string for a local postgre db and then we can simply switch the connection string for the AddDbContext.

https://docs.microsoft.com/en-us/ef/core/managing-schemas/migrations/providers?tabs=dotnet-core-cli

chrisK00:
… if you want to run the db with docker and npgsql

I tried following https://docs.microsoft.com/en-us/ef/core/managing-schemas/migrations/providers?tabs=dotnet-core-cli the second example (Using one context type) considering the first one would force us to change our codebase. I could not get it working unfortunately. The suggested solution here did nhttpsot work: ://github.com/dotnet/efcore/issues/14399

That is why i did the simplest thing possible, a on and off switch in the code. Managing migrations will be a challenge but at least we can work on the project now without docker. If anyone has a solution on the "using one context type" from the docs you are more than welcome to modify this branch.

closes #164


HellHunterMax (approved):
Great easy way to choose what to test the project with!

PR #173, opened 2021-09-15
Added authorisation logic to front end

### LINKED ISSUE #171: Add business logic to login components

PilgrimViis:
Finally, I'm done. I've changed the issue because the pull request is big enough and if registration will be in another one it'll be more convenient for a reviewer.

Shortlist of changes:
Added profile page for login testing purposes, which definitely requires a touch of a design. Added auth-interceptor to pass token towards backend with requests.


Client/src/app/_services/account.service.ts
```diff
@@ -26,12 +49,48 @@ export class AccountService {
...
+  }
+
+  private setCurrentUser(user: User) {
     this.currentUser.next(user);
   }
 
-  logout() {
-    localStorage.removeItem('user');
-    this.currentUser.next(undefined);
+  private handleError(error: HttpErrorResponse, showError: (message: string) => void) {
```
HellHunterMax: Backend returns a 400 when user is not found.

Server/JokesOnYou.Web.Api/appsettings.json
```diff
@@ -1,6 +1,6 @@
 {
   "ConnectionStrings": {
-    "PostgresConnectionString": "User ID=jokeuser;Password=password;Server=db;Port=5432;Database=jokesdb;Integrated Security=true;Pooling=true"
+    "PostgresConnectionString": "User ID=jokeuser;Password=password;Server=localhost;Port=5432;Database=jokesdb;Integrated Security=true;Pooling=true"
```
HellHunterMax: for me this line did not work.
i changed it back to "User ID=jokeuser;Password=password;Server=db;Port=5432;Database=jokesdb;Integrated Security=true;Pooling=true"
then it worked again.

HellHunterMax (approved):
seems to work fine. some edits were needed for it to run on my machine though.
The Joke details page seemed not to load.
I had to change the route to it from the jokes page so inside joke-card.component change the route to <a [title]="joke.premise" [routerLink]="['/joke/', joke.id]"> 
then that should work as well.

I also had do change the environments apiUrl: 'http://localhost:8000/api/'

Don't know if some of these issues are just mine though.

Client/src/app/_services/account.service.ts
chrisK00: We should also make use of interceptors, https://angular.io/api/common/http/HttpInterceptor that way we can centralize our exception logic and if needed navigate the user to a "error page". This might be a good way to introduce a package such as toastr for some smooth notifications

Server/JokesOnYou.Web.Api/Exceptions/WrongCredentialsException.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+
+namespace JokesOnYou.Web.Api.Exceptions
+{
+    public class WrongCredentialsException : Exception
```
chrisK00: Im being quite picky here but naming this BadCredentialsException might be more appropiate

Server/JokesOnYou.Web.Api/Middlewares/ExceptionMiddleware.cs
```diff
@@ -36,6 +36,7 @@ public async Task InvokeAsync(HttpContext context)
                     AppException => (int)HttpStatusCode.BadRequest,
                     UserRegisterException => (int)HttpStatusCode.BadRequest, //consider using built in ArgumentException. 
                     UserLoginException => (int)HttpStatusCode.Unauthorized,
+                    WrongCredentialsException => (int)HttpStatusCode.Forbidden,
```
chrisK00: Shouldnt this be a 401 unauthorized?

chrisK00 (approved):
Good stuff, keep it going Viis!

Server/JokesOnYou.Web.Api/Middlewares/ExceptionMiddleware.cs
HellHunterMax: One Line above there is also the user login exception which could be used.

Client/src/app/_services/account.service.ts
PilgrimViis: @chrisK00 I like the idea. Done.

PR #176, opened 2021-09-17
1. Removed all traces of docker including the docker compose file.yml

### LINKED ISSUE #175: Bye Docker
Since docker is not adding any value to the project right now but rather introducing more complexity, less time for the actual project and less welcoming to devs it is best to remove it.

chrisK00:
2. removed sqllite package since its not being used anymore
3. configured sql server in order to provide a connection string that works for everyone
4. To use the project write update-database in order for the database to be created and then run like you would normally do


HellHunterMax (approved):
If we want to use SQL Server this is fine and approved.

PR #183, opened 2021-09-19
moved token key to appsettings.development. Refactored service extens…

### LINKED ISSUE #129: Move services from startup to the ServiceExtension
We have this in Startup class:  `services.ConfigureAppServices(_config);`
But some services are still registered inside startup such as:
       ```services.AddScoped<IUserService, UserService>();
            services.AddScoped<IAuthService, AuthService>();```

Lets move theese to the configure app services method.

### LINKED ISSUE #178: Move JWT auth key to another location
devenv_NwCAlMsKXV

Move this key to appsettings.development.json as of right now so each person dont have to set a secret. Update the TokenService to take in the configuration in order to load it from the .json file.

chrisK00:
…ions to not return the collection. Moved services from startup in to serviceExtensions

closes #129 and #178 
Since #178 was so short i included it in this PR as i was already modifying one of the classes it resides in.


Server/JokesOnYou.Web.Api/Services/JwtTokenService.cs
```diff
@@ -13,11 +14,10 @@ public class JwtTokenService : ITokenService
         private readonly JwtSecurityTokenHandler _tokenHandler;
         private readonly byte[] _key;
 
-        public JwtTokenService()
+        public JwtTokenService(IConfiguration config)
```
PilgrimViis: It's not a bid deal, but I think it's a bit of overhead to pass the whole configuration to the constructor when we only need a key.

Server/JokesOnYou.Web.Api/Extensions/ServiceExtensions.cs
```diff
@@ -45,11 +64,9 @@ public static IServiceCollection ConfigureAppServices(this IServiceCollection se
                     ValidateAudience = false
```
PilgrimViis, line 70: What is our Admin validation policy will be? I know that some people use "aud" claim for it. They put audience "user" or "admin" and use it as a role. Others, for example, use custom claim or rely solely on a scope's claim.

PilgrimViis (approved):

Server/JokesOnYou.Web.Api/Services/JwtTokenService.cs
chrisK00: We can ofc bind to a class but then we have one class with one property, is there another approach?

Server/JokesOnYou.Web.Api/Extensions/ServiceExtensions.cs
chrisK00: Every user gets claims depending on their roles and we use thoose combined with ASP.net [Authorize(Roles = )] alternatively make use of policies. We have not yet implemented any roles tho

Server/JokesOnYou.Web.Api/Extensions/ServiceExtensions.cs
```diff
@@ -15,8 +19,20 @@ namespace JokesOnYou.Web.Api.Extensions
...
         {
+            services.AddScoped<IUserService, UserService>();
+            services.AddScoped<ITokenService, JwtTokenService>();
+            services.AddScoped<ITagService, TagService>();
+
+            services.AddScoped<IUserRepository, UserRepository>();
+            services.AddScoped<IUnitOfWork, UnitOfWork>();
+            services.AddScoped<IJokesRepository, JokesRepository>();
+            services.AddScoped<IJokesService, JokesService>();
+            services.AddScoped<ITagRepository, TagRepository>();
```
HellHunterMax, line 37: Maybe a bit overboard but.. should we give the scopes their own method?

HellHunterMax (approved):
nice!

Server/JokesOnYou.Web.Api/Extensions/ServiceExtensions.cs
chrisK00: We can place all data related services in their own method

PR #184, opened 2021-09-25
Add: Jokes search by Filters;

### LINKED ISSUE #91: feature: Get jokes by Filter

HellHunterMax:
Easy to add new jokes by adding a new filter with the IFilter interface and adding it to the JokeFiltersDto and adding it in JokesService.

How about this?

Closes: #91

### DIFF
--- a/Server/JokesOnYou.Web.Api/Repositories/JokesRepository.cs
+++ b/Server/JokesOnYou.Web.Api/Repositories/JokesRepository.cs
@@ -8,6 +8,9 @@ using Microsoft.EntityFrameworkCore;
 using System.Collections.Generic;
 using System.Threading.Tasks;
 using JokesOnYou.Web.Api.Models.Response;
+using JokesOnYou.Web.Api.Models.Interfaces;
+using System.Linq;
+using System.Data;
 
 namespace JokesOnYou.Web.Api.Repositories
 {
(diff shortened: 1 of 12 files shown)


chrisK00:
Good idea but there's something very important we need to have a look at
We chose to do a sort of N-tier architecture in order to decouple the components from each other, which is another reason we choose repository pattern because we want to abstract away our data access layer.

Whats happening here is that we are providing an expression from another layer as if it were a generic repository pattern. 

This could be simplified in to 
 ```
 class JokesParams
    {
        public string Search { get; set; } //search for any joke that contains the text
        public string AuthorId { get; set; } //search for a specific author's joke
        public int MinLikes { get; set; }
        public int MaxLikes { get; set; } 
    }
```

File: Server/JokesOnYou.Web.Api/Repositories/JokesRepository.cs [hunk 1]
```diff
@@ -37,5 +40,19 @@ public JokesRepository(DataContext context, IMapper mapper)
             await _context.Jokes.ProjectTo<JokeReplyDto>(_mapper.ConfigurationProvider)
                                 .FirstOrDefaultAsync(j => j.Id == id);
         public void DeleteJoke(Joke joke) => _context.Jokes.Remove(joke);
+
+        public async Task<IEnumerable<JokeReplyDto>> GetFilterdJokesAsync(IEnumerable<IFilter> filters)
+        {
+            var jokes = from j in _context.Jokes
```
chrisK00, line 46: We are not using query syntax in other places so might be a good idea to be consistent

chrisK00 (changes requested):

HellHunterMax:
@chrisK00 Thanks for the comments I agree with them and I'll have a look tonight! thanks! learnt a lot from this though!

PR #185, opened 2021-09-25
Tests for registration component added

### LINKED ISSUE #135: Home component for login/register
Look at the Figma mockups.

We want the client app to be as minimized as possible. Create a home component where the user can Login or Register. Make use of NgIf in order to only show one form at a time.  

[NgForms](https://angular.io/api/forms/NgForm) are just enough.

PilgrimViis:
I've added registration component logic and added a bit of testing.


Client/src/app/components/auth/registration/registration.component.html
```diff
@@ -91,7 +94,7 @@ <h1 class="offset-md-6 offset-3">Register</h1>
       Register
     </button>
     <span class="col-md-3">
-      Have already an account?&nbsp;<a href="/login" class="underlineLink">Go to login</a>      
+      Have already an account?&nbsp;<a href="/login" class="underlineLink">Go to login</a>
```
HellHunterMax, line 97: This should read: Already have an acount?

HellHunterMax (commented):
Maybe some kind of Message like Account created Successfully ?
And i just saw the link back to the jokes does not work anymore.
@chrisK00 should look at this. but i just checked if it works and it does ;)

good job imo its really easy to follow even for someone who does not know angular ;)

Client/src/app/components/auth/registration/registration.component.html
PilgrimViis: > And i just saw the link back to the jokes does not work anymore.

What link? /jokes?

HellHunterMax: Sorry I should have been more clear. The left  button on the navbar does not link back to jokes.

Server/JokesOnYou.Web.Api/Models/Request/UserRegisterDTO.cs
```diff
@@ -8,9 +9,13 @@ public class UserRegisterDto
         [EmailAddress]
         public string Email { get; set; }
         [Required]
+        [MinLength(4)]
         public string UserName { get; set; }
         [Required]
         public string Password { get; set; }
-
+        [Required]
```
chrisK00: Doesnt have to be required, can be false by default. Should we also check that the user is above a specific age?

Niisc: We probably should, i dont think its legal to show nsfw stuff to people under 18 or something

HellHunterMax: nsfw does not mean Sexy stuff per se and we wont even allow like naked stuff. (not that we have pictures yet)
Should just be there for like Swear words and stuff? so no date required for that.

Client/src/app/components/auth/registration/registration.component.html
PilgrimViis: > 
> 
> Sorry I should have been more clear. The left button on the navbar does not link back to jokes.

No problem. Done.

Server/JokesOnYou.Web.Api/Models/Request/UserRegisterDTO.cs
PilgrimViis: Done

chrisK00 (approved):

Issue #189, opened 2021-09-27
Refactor: Change DateTime instances


HellHunterMax:
`This is an easy issue for people new to the project wanting to learn it or new to programming.`

#### "I'd recommend using `DateTime.Now` whenever you're displaying a date to a human being."
vs
## example:
Right now: Jokes saves to the Database Using DateTime.Now, this should be UtcNow.

## What to do:
Find all instances of DateTime and figure out which ones belong where.

## Source:
[datetime.now vs datetime.utcnow](https://stackoverflow.com/questions/62151/datetime-now-vs-datetime-utcnow)

ysnarafat:
Looks interesting, want to work on this issue.

HellHunterMax:
> Looks interesting, want to work on this issue.

Great if you need any help just call out on discord!

ysnarafat:
The discord invitation link shows invalid. Will you place the invitation link here?

chrisK00:
> The discord invitation link shows invalid. Will you place the invitation link here?

https://discord.gg/F3Z9EFadP5

PR #192, opened 2021-10-06
Change corresponding DateTime instances

ysnarafat:
Change DateTime instance for issue #189


chrisK00 (approved):
Nothing to comment on, looks good 👍

PR #193, opened 2021-10-22
Add pagination for GetAllUser

### LINKED ISSUE #187: Feature: add paging to Users.
Right now when you Call the get Users  you get a list of all Users.
this needs to be limited with pages.

DONT USE THIS EXACTLY BUT IT MIGHT HELP:
https://docs.microsoft.com/en-us/aspnet/mvc/overview/getting-started/getting-started-with-ef-using-mvc/sorting-filtering-and-paging-with-the-entity-framework-in-an-asp-net-mvc-application

ysnarafat:
closes #187


Server/JokesOnYou.Web.Api/Controllers/UserController.cs
```diff
@@ -7,10 +7,11 @@
 using System.Collections.Generic;
 using System.Threading.Tasks;
 using JokesOnYou.Web.Api.Models.Response;
+using JokesOnYou.Web.Api.Models.Request.Query;
 
 namespace JokesOnYou.Web.Api.Controllers
 {
-    [Authorize]
+    //[Authorize]
```
chrisK00: Why was this removed?

Server/JokesOnYou.Web.Api/Extensions/ServiceExtensions.cs
```diff
@@ -41,6 +42,10 @@ public static void ConfigureDataAccess(this IServiceCollection services, IConfig
             {
                 options.UseSqlServer(config.GetConnectionString("SQLserverConnection"));
             });
+            services.AddDbContext<IdentityDbContext<User>>(options =>
```
chrisK00: The db has already been configured, what does this do?

Server/JokesOnYou.Web.Api/Helpers/PaginatedList.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace JokesOnYou.Web.Api.Helpers
+{
+    public class PaginatedList<T> : List<T>
+    {
+        public int CurrentPage { get; set; }
+        public int PageSize { get; set; }
+        public int NextPage { get; set; }
+        public int PrevousPage { get; set; }
+        public int PageNumber { get; set; }
```
chrisK00: Is PageNumber not the same as CurrentPage? Or was this maybe supposed to be total pages considering the client needs to know how many pages we have. Probably want to have the amount of items aswell so the client can display something like:
found 1000 items
page 1/30

Server/JokesOnYou.Web.Api/Helpers/PaginatedList.cs
```diff
@@ -0,0 +1,39 @@
...
+        public PaginatedList(List<T> items, int pageNumber, int pageSize)
```
chrisK00: Can be an IEnumerable

Server/JokesOnYou.Web.Api/Helpers/PaginatedList.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            PageSize = pageSize;
+            CurrentPage = pageNumber;
+
+            AddRange(items);
+        }
+
+        public static async Task<PaginatedList<T>> ToPaginatedList(IQueryable<T> source, int pageNumber, int pageSize, Expression<Func<T,bool>> expression)
```
chrisK00: naming - ToPaginatedListAsync
the expression should not be sent here since the IQueryable is already passed down, we can do the filtering in our repository

Server/JokesOnYou.Web.Api/Models/Request/Query/PaginationQuery.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace JokesOnYou.Web.Api.Models.Request.Query
+{
+    public abstract class PaginationQuery : IValidatableObject
```
chrisK00: Does not have to be abstract, sometimes we might want to use Pagination and not filtering, pagination is always required tho.

I would suggest changing the name of the class to something like PaginationParams since this is just a POCO and not directly tied to a query

Server/JokesOnYou.Web.Api/Models/Request/Query/PaginationQuery.cs
```diff
@@ -0,0 +1,44 @@
...
+            {
+                return _pageSize;
+            }
+            set
+            {
+                _pageSize = (value > maxPageSize) ? maxPageSize : value;
+            }
+        }
+
+        public IEnumerable<ValidationResult> Validate(ValidationContext validationContext)
```
chrisK00: We could add data annotations or simply set default values in case they send weird numbers in the query

Server/JokesOnYou.Web.Api/Models/Request/Query/UserPaginationQueryParameters.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace JokesOnYou.Web.Api.Models.Request.Query
+{
+    public class UserPaginationQueryParameters : PaginationQuery
```
chrisK00: Can be shortened to UserParameters or UserParams whatever suits you

Server/JokesOnYou.Web.Api/Repositories/Interfaces/IUserRepository.cs
```diff
@@ -13,7 +14,7 @@ public interface IUserRepository
         Task<User> CreateUserAsync(UserRegisterDto userRegisterDTO);
         Task DeleteUserAsync(User user);
         Task<UserReplyDto> GetUserReplyAsync(string id);
-        Task<IEnumerable<UserReplyDto>> GetUsersReplyDtoAsync();
+        Task<PaginatedList<User>> GetAllUserAsync(UserPaginationQueryParameters parameters);
```
chrisK00: The naming convention and return type was changed here from
 Task<IEnumerable<UserReplyDto>> GetUsersReplyDtoAsync();
 should become
 Task<PaginatedList<UserReplyDto>> GetUsersReplyDtoAsync

Server/JokesOnYou.Web.Api/Repositories/UserRepository.cs
```diff
@@ -65,8 +70,15 @@ await _userManager.Users.ProjectTo<UserReplyDto>(_mapper.ConfigurationProvider)
 
         public async Task<User> GetUserByEmailAsync(string email) => await _userManager.FindByEmailAsync(email);
 
+        public async Task<PaginatedList<User>> GetAllUserAsync(UserPaginationQueryParameters parameters)
+        {
+            var result =  await PaginatedList<User>.ToPaginatedList(
+                _dbContext.Users.AsNoTracking(),
```
chrisK00: Curious: why are we using the dbcontext instead of UserManager here?

Server/JokesOnYou.Web.Api/Services/Interfaces/IUserService.cs
```diff
@@ -13,7 +14,7 @@ public interface IUserService
         Task<UserReplyDto> GetUserReplyById(string id);
         Task UpdateUser(UserUpdateDto user);
         //Task<User> Authenticate(string username, string password);
-        Task<IEnumerable<UserReplyDto>> GetAll();
+        Task<IEnumerable<UserReplyDto>> GetAll(UserPaginationQueryParameters parameters);
```
chrisK00: Should return the paginatedList so we can append the headers containing the pagination information

Server/JokesOnYou.Web.Api/Services/UserService.cs
```diff
@@ -29,9 +33,11 @@ public UserService(IUserRepository userRepo, IMapper mapper, IUnitOfWork unitOfW
             _tokenService = tokenService;
         }
 
-        public async Task<IEnumerable<UserReplyDto>> GetAll()
+        public async Task<IEnumerable<UserReplyDto>> GetAll(UserPaginationQueryParameters parameters)
         {
-            return await _userRepository.GetUsersReplyDtoAsync();
+            var users = await _userRepository.GetAllUserAsync(parameters);
+
+            return users.AsQueryable().ProjectTo<UserReplyDto>(_mapper.ConfigurationProvider);
```
chrisK00: This is a job for the repository. The repository is what is between our database provider and our application, it makes us less dependent on who is providing us with the data, this way we are breaking the repository pattern so we need to move all code that has to do with the db or EF in to the repository, the service is simply a service that is a layer made to make our controllers clean

chrisK00 (changes requested):

Server/JokesOnYou.Web.Api/Controllers/UserController.cs
ysnarafat: Ah, forgot to uncomment after testing

Server/JokesOnYou.Web.Api/Helpers/PaginatedList.cs
ysnarafat: Yeah, I messed up somehow.

chrisK00: > Yeah, I messed up somehow.
No worries, if you have 0 messups then what are reviews for 😄

Server/JokesOnYou.Web.Api/Controllers/UserController.cs
chrisK00: This is still uncommented

Server/JokesOnYou.Web.Api/Controllers/UserController.cs
```diff
@@ -1,16 +1,17 @@
...
+using JokesOnYou.Web.Api.Models.Request.Query;
+using JokesOnYou.Web.Api.Models.Response;
 using JokesOnYou.Web.Api.Services.Interfaces;
-using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Logging;
-using System.Collections.Generic;
-using System.Threading.Tasks;
-using JokesOnYou.Web.Api.Models.Response;
+using Newtonsoft.Json;
```
chrisK00: System.Text.Json is already installed so we dont need two json serializers

Server/JokesOnYou.Web.Api/Models/Request/Query/PaginationParameters.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Threading.Tasks;
+
+namespace JokesOnYou.Web.Api.Models.Request.Query
+{
+    public class PaginationParameters //: IValidatableObject
+    {
+        const int maxPageSize = 50;
+        private int _pageSize = 10;
+
+        [RegularExpression(@"^\d*[1-9]\d*$", ErrorMessage = "PageNumber cannot be zero or negative")]
```
chrisK00: There is a range data annotation so we dont have to use regular expressions

Server/JokesOnYou.Web.Api/Repositories/UserRepository.cs
```diff
@@ -65,8 +68,23 @@ await _userManager.Users.ProjectTo<UserReplyDto>(_mapper.ConfigurationProvider)
 
         public async Task<User> GetUserByEmailAsync(string email) => await _userManager.FindByEmailAsync(email);
 
+        public async Task<PaginatedList<UserReplyDto>> GetUsersReplyDtoAsync(UserParameters parameters)
+        {
+            var users = _userManager.Users.AsNoTracking();
+
+            if (string.IsNullOrEmpty(parameters.SearchText) != true ||
```
chrisK00: this code can be written like this:
!string.IsNullOrEmpty(parameters.SearchText)
or
string.IsNullOrEmpty() == false

chrisK00 (changes requested):
Great improvments, almost there

chrisK00 (approved):
Looks great!🥳 Well done

ysnarafat:
> Looks great!🥳 Well done

Thanks for the thorough review.

HellHunterMax (approved):
Well done cool way of doing this!

PR #197, opened 2021-11-09
Fixed so we only make 2 queries to the db for getting jokes with auth…

### LINKED ISSUE #196: Reduce the amount of queries made to the database and only select the necessary data
### The problem in JokeService
1. For every joke that has an author, the author will be loaded. An author is retrieved n times because we don't cache the author since we are using .AsNoTracking
2. We load the whole user entity event though we only want the username

### The solution in JokesRepository
1. Inside GetJokeDtosAsync we know that this will return a JokeReplyDto which has an author entity, that means we can directly populate this here.
2. We can start defining the query ` var query = _context.Jokes.AsNoTracking(); // asnotracking will make it a queryable`
3. We do our filters on the query such has ```if (!string.IsNullOrWhiteSpace(jokesFilter.AuthorId))
            {
                query = query.Where(joke => joke.AuthorId == jokesFilter.AuthorId);
            }```
4. We paginate the query which means we need to refactor our PaginatedList to return the iqueryable instead of executing it so we later on can execute the .ToPaginatedList `var jokes = await query.Paginate(int pageSize)`
5. Now we need to load the relatedAuthors and we can use the query to only make one request to the db by doing so we are making a `SELECT `subquery instead of a `Join `which becomes more performant.  ```var authors = await _context.Users.Where(u => query.Select(j => j.AuthorId).Contains(u.Id))
                .Select(x => new { x.UserName, x.Id }).ToListAsync()```
6. Almost done, just need to create the ReplyDto ```return jokes.Select(j => new JokeReplyDto
            {
                Author = new JokeAuthorDto
                {
                    Id = j.AuthorId,
                    UserName = authors.FirstOrDefault(x => x.Id == j.AuthorId).UserName
                },
            });```

### What have we changed?
- We only load the data necessary by doing multiple Select statements
- We make a query, filter, paginate and use this query in a query where we want to get related data. That means we always only make 2 Queries to the database
- We Map our results to a Dto

chrisK00:
…or. Refactored the ToPaginatedList so that we can manually append pagination to the query and that means we need to get the count first.

If you have any questions about the code don't hesitate to ask.

closes #196

### DIFF
--- a/Server/JokesOnYou.Web.Api/Extensions/QueryableExtensions.cs
+++ b/Server/JokesOnYou.Web.Api/Extensions/QueryableExtensions.cs
@@ -1,0 +1,27 @@
+﻿using Microsoft.EntityFrameworkCore;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace JokesOnYou.Web.Api.Extensions
+{
+    public static class QueryableExtensions
+    {
+        /// <summary>
+        ///  Gets the total amount of items in the database and then applies pagination to the query
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="source"></param>
+        /// <param name="pageNumber"></param>
+        /// <param name="pageSize"></param>
+        /// <returns>The query paginated and the count before pagination was applied</returns>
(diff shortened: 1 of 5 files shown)


HellHunterMax (approved):
Yup this works. Tested with 16 jokes and it makes 3 calls to the DB
1 for the count.
1 for the users
1 for the jokes.

PR #211, opened 2021-11-22
Implement GetTags using a JokeId

### LINKED ISSUE #203: Feature: GetTags using a JokeId
the Website loads the jokes and then asks the backend for the tags associated with the joke.

use the UserTagJoke Model to find the tags Tagged with the Joke.

talatmorshedtanbir:
closes #203


HellHunterMax (changes requested):
Great job!
I like what you did with the query.

One issue though, we are returning the number of likes on the whole Tag and not the likes for that joke. so it should get the likes from the UserJokeTag and place those on the TagRelplyDto.

talatmorshedtanbir:
Modified the query. check please

HellHunterMax (approved):
Perfect I think.

